PR #6, opened 2018-03-23
Upgrade base

carver:
## What was wrong?

Was missing some upgrades from the template

## How was it fixed?

Merge in template upgrades, like:

- add doctest
- add pypy3 tests
- readme improvements
- deployment standardization

#### Cute Animal Picture

### DIFF
--- a/eth_account/datastructures.py
+++ b/eth_account/datastructures.py
@@ -15,3 +15,14 @@ class AttributeDict(AttrDict):
             'This data is immutable -- create a copy instead of modifying. '
             'For example, AttributeDict(old, replace_key=replace_val).'
         )
+
+    def _repr_pretty_(self, builder, cycle):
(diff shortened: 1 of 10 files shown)


carver:
I pushed the branch upstream, then deleted it, so the travis push test failed. As long as the PR test and circle tests succeed, this is good to go.

eth_account/datastructures.py
```diff
@@ -15,3 +15,14 @@ def __setitem__(self, attr, val):
             'This data is immutable -- create a copy instead of modifying. '
             'For example, AttributeDict(old, replace_key=replace_val).'
         )
+
+    def _repr_pretty_(self, builder, cycle):
```
pipermerriam: Awesome.  I looked for this API once and couldn't find it.

PR #7, opened 2018-03-23
doc: Use `getpass` instead of inline passphrase.

veox:
## What was wrong?

Documentation shows passphrase in cleartext.

## How was it fixed?

Provide a quick hint (`getpass.getpass()`) on how this can be avoided.

(Since the examples show a REPL, IMO it's fine using an interactive prompt.)

#### Cute Animal Picture

> "You what in there?.."

lemur with wide-open eyes

Source: [Leesburg animal park](https://www.leesburganimalpark.com/meet-the-animals/)

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -113,7 +113,8 @@ class Account(object):
              'id': 'b812f3f9-78cc-462a-9e89-74418aa27cb0',
              'version': 3}
 
-            >>> Account.decrypt(encrypted, 'correcthorsebatterystaple')
+            >>> import getpass
+            >>> Account.decrypt(encrypted, getpass.getpass())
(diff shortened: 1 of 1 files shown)


carver:
Cool, thanks for this.

My only concern here is that I'd like to get some/all of these snippets running with `doctest`. So I'd like to think through forward compatibility. I suppose we could install a mocked version of `getpass` during setup. Any other ideas?

veox:
Well, could intead recommend in-text of the apidoc, instead of example?..

carver:
I like having the doctests run against as many examples as possible, so let's just use `getpass` for now. When someone gets around to enabling doctests, they'll just have to figure out a way around it.

Thanks!

PR #16, opened 2018-04-23
Update for rlp 1.0.0

pipermerriam:
## What was wrong?

Updating to be compatible with the latest `rlp==1.0.0`

## How was it fixed?

Removed use of `exclude` API.

#### Cute Animal Picture

baby-camel-face


carver:
Sweet

PR #23, opened 2018-05-23
Implement __eq__ and __hash__ for LocalAccount

### LINKED ISSUE #21: Implement LocalAccount equality test

```py
from eth_account import Account
key = b'a' * 32
acct1 = Account.privateKeyToAccount(key)
acct2 = Account.privateKeyToAccount(key)
assert acct1 == acct2, "local accounts with the same private key should evaluate as equal"
```

### Full error output 

```sh
AssertionError: local accounts with the same private key should evaluate as equal
```

## How can it be fixed?

Add an `__eq__` to `LocalAccount`, test that both objects are exactly the same type, and that they have the same key.

Bonus:  Make the private key attribute of `LocalAccount` immutable, and implement `__hash__` as `hash(<private_key>)`.

stefanmendoza:
## What was wrong?

Fixes #21 

## How was it fixed?

Implemented `__eq__` and `__hash__` for `LocalAccount` + new tests.

#### Cute Animal Picture

### DIFF
--- a/eth_account/local.py
+++ b/eth_account/local.py
@@ -29,9 +29,16 @@ class LocalAccount(object):
         self.address = key.public_key.to_checksum_address()
 
         key_raw = key.to_bytes()
-        self.privateKey = key_raw
+        self._privateKey = key_raw
 
         self._key_obj = key
+
+    @property
+    def privateKey(self):
+        '''
+        Get the private key.
+        '''
+        return self._privateKey
 
     def encrypt(self, password):
         '''
@@ -59,3 +66,9 @@ class LocalAccount(object):
 
     def __bytes__(self):
         return self.privateKey
+
+    def __hash__(self):
+        return hash(self.privateKey)
+
+    def __eq__(self, other):
+        return (isinstance(other, type(self))) and (self.privateKey == other.privateKey)
--- a/tests/core/test_accounts.py
+++ b/tests/core/test_accounts.py
@@ -82,6 +82,13 @@ def test_eth_account_create_variation(acct):
     assert account1 != account2
 
 
+def test_eth_account_equality(acct):
+    key = b'a' * 32
+    acct1 = acct.privateKeyToAccount(key)
+    acct2 = acct.privateKeyToAccount(key)
+    assert acct1 == acct2
+
+
 def test_eth_account_privateKeyToAccount_reproducible(acct, PRIVATE_BYTES):
     account1 = acct.privateKeyToAccount(PRIVATE_BYTES)
     account2 = acct.privateKeyToAccount(PRIVATE_BYTES)


eth_account/local.py
```diff
@@ -59,3 +59,9 @@ def signTransaction(self, transaction_dict):
 
     def __bytes__(self):
         return self.privateKey
+
+    def __hash__(self):
+        return hash(self.privateKey)
```
carver: Implementing `__hash__` implies that the object is immutable, so we need to make it harder to accidentally change the private key.

That could look something like: moving `privateKey` to `_privateKey`, and adding a `@property` for `privateKey`.

Or, if you don't feel like doing that, you can remove the `__hash__` implementation.

eth_account/local.py
```diff
@@ -59,3 +59,9 @@ def signTransaction(self, transaction_dict):
...
+
+    def __eq__(self, other):
+        return (isinstance(other, self.__class__)) and (self.__hash__ == other.__hash__)
```
carver: Instead of `self.__hash__ == other.__hash__`, try something like `self.privateKey == other.privateKey`.

pipermerriam: Also, I think `type(self)` is preferred over `self.__class__`.  Same with `hash(self)` over `self.__hash__`, largely to avoid private attribute access.

eth_account/local.py
stefanmendoza: Yeah, I was looking into doing the `_privateKey` approach for implicit private scope, but wasn't sure if implementing something like a getter was the approach we'd want as I'm assuming there are implementations utilizing `.privateKey` and I wanted the change to be passive. I've dabbled with Python, but not super extensively, so still finding my way around. Will look into the `@property` annotation. :)

eth_account/local.py
stefanmendoza: @carver  @pipermerriam  Thanks for the feedback! I'll look into getting this switched.

eth_account/local.py
carver: Ah, great. It's definitely worth a look. Something like:

```py
@property
def privateKey(self):
    return self._privateKey
```

That approach, by default, will prevent any callers from modifying the account with `acct.privateKey = some_new_key`, but still give access to use `extracted_key = acct.privateKey`.

stefanmendoza:
@carver okay, I think I'm ready to get this actually reviewed! Since equality was tested through [this test](https://github.com/ethereum/eth-account/blob/master/tests/core/test_accounts.py#L79-L82) by using [`Account.create()`]https://github.com/ethereum/eth-account/blob/master/eth_account/account.py#L64) to generate a random key, I made a new test utilizing that method and doing a deepcopy of the object. Let me know if that's sufficient for testing this change. Also implemented the changes suggest by you and @pipermerriam. Thanks again for the feedback!

Issue #25, opened 2018-06-04
Create `LedgerAccount` type


pipermerriam:
### What is wrong

Hardware wallets are great!

### How can they be implemented.

https://gist.github.com/bargst/5f896e4a5984593d43f1b4eb66f79d68

This appears to be a great starting point.

The base `Account` object should probably assume the *default* path.  We should probably also expose an API for accessing sub-accounts at a different path assuming the ledger supports arbitrary BIP32 paths.

pipermerriam:
Probably not a blocker but: #24 is likely related to this in some way since both leverage HD wallets in some form.

bargst:
Some work merged in a local feature branch in #34 pending in-depth review and security audit.

pipermerriam:
Alternate implementation from the shadowlands guy

https://github.com/kayagoban/shadowlands-core/blob/master/shadowlands/credstick/ledger_ethdriver.py

benjyz:
@pipermerriam has anyone attempted any progress in this direction? digging into this a bit

this core moved to 
https://github.com/kayagoban/shadowlands-core/blob/master/shadowlands/credstick/ledger_ethdriver.py

https://github.com/kayagoban/shadowlands/blob/master/shadowlands/credstick/ledger_ethdriver.py

this gist needs rebase
https://gist.github.com/bargst/5f896e4a5984593d43f1b4eb66f79d68

benjyz:
made some progress on python with ledger using #34 branch

https://gist.github.com/benjyz/c4d96eb79a46c7d3d34d64f4870e2f8d

managed to get the version number and address from the device. have to read up on derivation paths. for some reason ledger live uses different paths

the code uses:
ETH_DERIVATION_PATH_PREFIX = "m/44'/60'/0'/"
ledger live uses 
ETH_DERIVATION_PATH_PREFIX = "44'/60'/1'/0'/0"

benjyz:
should be this

ETH_DERIVATION_PATH = "m/44'/60'/0'/0/0"

benjyz:
maybe add a flag like this
https://github.com/ethereum/eth-account/blob/master/eth_account/account.py#L79

PR #26, opened 2018-06-05
Ledger support

bargst:
## What was wrong?

Issue #25 

## How was it fixed?

Create a class LedgerAccount

#### Cute Animal Picture


eth_account/ledger.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        path = ETH_DERIVATION_PATH_PREFIX + str(offset)
+        result = bytes()
+        elements = path.split('/')
+        for pathElement in elements:
+            element = pathElement.split("'")
+            if len(element) == 1:
+                result = result + struct.pack(">I", int(element[0]))            
+            else:
+                result = result + struct.pack(">I", 0x80000000 | int(element[0]))
```
pipermerriam: Any chance we can get an explanation of what this code is doing and why with some inline comments?

eth_account/ledger.py
```diff
@@ -0,0 +1,124 @@
...
+        apdu += bytes([len(donglePath) // 4])
+        apdu += donglePath
+
+        result = self.dongle.exchange(apdu, timeout=60)
+
+        # Parse result
+        offset = 1 + result[0]
+        address = result[offset + 1 : offset + 1 + result[offset]]
+
+        return f'0x{address.decode()}'
```
pipermerriam: This should use `eth_utils.to_normalized_address` (assuming `address` is a bytes value.

eth_account/ledger.py
```diff
@@ -0,0 +1,124 @@
...
+        elements = path.split('/')
+        for pathElement in elements:
+            element = pathElement.split("'")
+            if len(element) == 1:
+                result = result + struct.pack(">I", int(element[0]))            
+            else:
+                result = result + struct.pack(">I", 0x80000000 | int(element[0]))
+        return result
+
+    def get_addresses(self, offset):
```
pipermerriam: The name of this function suggests that it will return multiple addresses, but it seems to only return a single one.

eth_account/ledger.py
```diff
@@ -0,0 +1,124 @@
...
+
+        return f'0x{address.decode()}'
+
+    def get_offset(self, address):
+        """
+        Convert an address to the HD wallet tree offset
+        """
+        offset = 0
+        while address != self.get_addresses(offset):
+            offset += 1
```
pipermerriam: Instead of this mutation pattern, how about:

```python
import itertools

for offset in itertools.count():
    ...
```

Also, lets set an upper bound on this to some reasonable value (like 20?).

carver: > Also, lets set an upper bound on this to some reasonable value (like 20?).

I'm not sure we can define what maximum number of addresses should be allowed against a wallet. I guess we can start with something like: how many addresses can we check in one second? At the least, we would need an escape hatch for people who might have tons of addresses, like:

```py
    def get_offset(self, address, maximum_search_length=100):
```

pipermerriam: Good plan, like the kwarg.  how about `search_limit`?  We should probably also allow for specifying the starting offset value.

bargst:
Thanks for the feedback, I updated accordingly ! I will probably change the offset name to something more explicit like _account_id_.

Next is signTransaction() then encrypt() and signHash()

pipermerriam:
@carver probably something for us to do separately, but this feels like the time to create an abstract base class for accounts so the public *required* API is explicitly defined.

carver:
Note that `encrypt()` is probably not possible to implement (I hope!), so just the other two are sufficient. A draft of the spec is in #28.

eth_account/ledger.py
```diff
@@ -0,0 +1,193 @@
...
+        return result
+
+    def _send_to_device(self, apdu):
+        try:
+            if self.device is None:
+                self.device = ledgerblue.comm.getDongle(debug=False)
+
+            reply = self.device.exchange(apdu, timeout=60)
+        except:
+            raise Exception('Something went wrong when communicating with device')
```
dylanjw: Im not too keen on masking the underlying exception.

carver:
After a rebase on master, you can subclass `BaseAccount`.

eth_account/ledger.py
bargst: Ok, I will remove the try catch

bargst:
@carver as defined, signHash() is not doable with the ledger, because what is sent to the device is not the hash of the message, it is the whole message. The message is then hashed directly in the ledger and vrs signatur are then returned. 

Is it possible to have the message passed to signHash() in BaseAccount ?

bargst:
@pipermerriam @carver I consider dev done, could you review ? 
What's missing is commit history clean-up and a decision about signHash() issue.

carver:
> decision about signHash() issue.

Yeah, signing is ugly right now. The process for hashing messages is not well standardized. It's baked into the hardware keys, so we can't have a single `signMessage` implemented by all the different signers as a drop-in replacement. They all do it a little differently.

So to support ledger signing, it should probably just be a custom method. `ledgerSignMessage()` is an option for the name. Alternatively, you could name it `defunctSignMessage()` (if it's the exact same hashing process as `geth`). It's [called the `defunct` style elsewhere in eth-account](http://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.messages.defunct_hash_message).

You can raise a `NotImplementedError` from `signHash` saying that the Ledger doesn't support signing opaque hashes.

> What's missing is commit history clean-up and a decision about signHash() issue.

Also, passing CI tests :)

dylanjw:
Because ledgerblue depends on pycrypto, and some of the other dependencies (eth-keyfile) have replaced it with pycryptodome, we are getting errors on this branch, like the following:

```
ImportError while importing test module '/home/dwilson/Develop/Projects/ethereum/eth-account/tests/core/test_ledger.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/core/test_ledger.py:5: in <module>
    from eth_account import (
eth_account/__init__.py:1: in <module>
    from eth_account.account import Account  # noqa: F401
eth_account/account.py:10: in <module>
    from eth_keyfile import (
venv/lib/python3.6/site-packages/eth_keyfile/__init__.py:7: in <module>
    from eth_keyfile.keyfile import (  # noqa: F401
venv/lib/python3.6/site-packages/eth_keyfile/keyfile.py:6: in <module>
    from Crypto.Cipher import AES
venv/lib/python3.6/site-packages/Crypto/Cipher/__init__.py:3: in <module>
    from Crypto.Cipher._mode_ecb import _create_ecb_cipher
venv/lib/python3.6/site-packages/Crypto/Cipher/_mode_ecb.py:29: in <module>
    from Crypto.Util._raw_api import (load_pycryptodome_raw_lib,
venv/lib/python3.6/site-packages/Crypto/Util/_raw_api.py:32: in <module>
    from Crypto.Util.py3compat import byte_string
E   ImportError: cannot import name 'byte_string'
```

There is an open issue on ledgerblue: https://github.com/LedgerHQ/blue-loader-python/issues/44

carver:
Hm, maybe we should use `pycryptodomex` in `eth-hash`, `eth-keys`, `eth-keyfile`, etc, as recommended here: http://pycryptodome.readthedocs.io/en/latest/src/introduction.html

carver:
Right now the tests are failing for fixable reasons, so I want to make sure those get resolved before we hide them in a test that's skipped by CI.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+    Ledger Ethereum App Protocol Spec is located at:
+    <https://github.com/LedgerHQ/blue-app-eth/blob/master/doc/ethapp.asc>
+
+    References:
+    - https://github.com/LedgerHQ/blue-app-eth/blob/master/getPublicKey.py
+    - https://github.com/bargst/pyethoff/blob/master/tx_sign.py
+
+    '''
+
+    def __init__(self, account_id=0, address='', path_prefix=ETH_DERIVATION_PATH_PREFIX):
```
pipermerriam: I'd prefer `None` as the default value for `address` here.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+        '''
+        Convert an address to the HD wallet tree account_id
+        Start search at an account_id. This would allow to search deeper if required.
+        Default search_limit at 20 take about 5s to reach.
+        '''
+        address = eth_utils.to_normalized_address(address)
+
+        for account_id in itertools.count(start=search_account_id):
+            if account_id > search_limit:
+                raise Exception(f'Address {address} not found' +
```
pipermerriam: Prefer use of `ValueError` or even a custom exception type here rather than a bare `Exception`

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+                                f'(search_limit={search_limit}, ' +
+                                f'search_account_id={search_account_id})')
+            if address == self.get_address(account_id):
```
pipermerriam: This should probably use `eth_utils.is_same_address`

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+                return account_id
+
+    def list(self, limit=5, page=0):
+        '''
+        List Ethereum HD wallet adrress of the ledger device
+        '''
+        return list(map(lambda account_id: self.get_address(account_id),
```
pipermerriam: I think the  lambda isn't needed here, you should be able to just pass in `self.get_address`.

Alternate way to do this which I think is more readable

```python
return [self.get_address(account_id) for account_id in range(...)]
```

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+
+        for account_id in itertools.count(start=search_account_id):
+            if account_id > search_limit:
+                raise Exception(f'Address {address} not found' +
+                                f'(search_limit={search_limit}, ' +
+                                f'search_account_id={search_account_id})')
+            if address == self.get_address(account_id):
+                return account_id
+
+    def list(self, limit=5, page=0):
```
pipermerriam: Alternate name suggestion `get_addresses`

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+        List Ethereum HD wallet adrress of the ledger device
+        '''
+        return list(map(lambda account_id: self.get_address(account_id),
+                        range(page * limit, (page + 1) * limit)))
+
+    def signHash(self, message_hash):
+        '''
+        Not available with a Ledger
+        '''
+        raise Exception('signHash() is not available within ledger devices')
```
pipermerriam: Can you use `NotImplementedError` here.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # Sign with dongle
+        result = self._send_to_device(apdu)
+
+        # Retrieve VRS from sig
+        v = result[0]
+        r = int.from_bytes(result[1:1 + 32], 'big')
+        s = int.from_bytes(result[1 + 32: 1 + 32 + 32], 'big')
+
+        rlp_encoded = encode_transaction(unsigned_transaction, vrs=(v, r, s))
```
pipermerriam: I'm curious if we should have a bit of validation here to sanity check that the returned signature does indeed belong to the expected signing address?

setup.py
```diff
@@ -6,6 +6,9 @@
 )
 
 extras_require={
+    'ledger': [
+        "ledgerblue==0.1.17",
```
pipermerriam: Should we change this to be a version range, maybe just pinned within the `minor` version?  `ledgerblue>=0.1.17,<0.2.0`

carver:
> look at the complexity to just extract the dongle interface to the device from ledgerblue to import it in eth-account directly

Sorry, I don't know much about the internals here of the dongle interface. Can you show how this proposal would change the eth-account API?

bargst:
@carver I see it as copying and adapting HIDDongleHIDAPI class from https://github.com/LedgerHQ/blue-loader-python/blob/master/ledgerblue/comm.py#L56 and inserting it in ledger.py directly. 
Then in LedgerAccount._send_to_device() use this new class instead of ledgerblue.comm.getDongle()

This HIDDongleHIDAPI class only has a dependency on 'hidapi>=0.7.99' and so we can get rid of ledgerbue requirement, but off course I need to make a small POC ...

pipermerriam:
> so we can get rid of ledgerbue requirement

This would be great in theory.  Will have to see how it looks in practice.

bargst:
I don't know why but I can't find a way to make the  [defunctSignMessage()](https://github.com/LedgerHQ/blue-app-eth/blob/master/doc/ethapp.asc#sign-eth-personal-message) work with my ledger and I don't think it is necessary for supporting ledger in `eth-account` so I won't dig deeper and instead focus on ldegerblue dependency.

bargst:
I will try to simplify USBDevice class to be cleaner. Maybe we can do like geth where they have implemented a USBDriver so it can be used by other hardware device: https://github.com/ethereum/go-ethereum/blob/master/accounts/usbwallet/ledger.go

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,399 @@
...
+        data = ""
+        while len(data) == 0:
+            data = self.device.read(65)
+            if not len(data):
+                if time.time() - start > timeout:
+                    raise Exception("Timeout")
+                time.sleep(0.0001)
+        return bytearray(data)
+
+    def wrapCommandAPDU(self, channel, command, packetSize):
```
pipermerriam: This method doesn't appear to use `self`.  What do you think about extracting it to a utility function so that it can have some basic tests written for it (which will also allow it to be easily tested in isolation)

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,399 @@
...
+                possibleCause = "Not enough space?"
+            if sw == 0x6484:
+                possibleCause = "Are you using the correct targetId?"
+            raise Exception("Invalid status %04x (%s)" % (sw, possibleCause))
+        return response
+
+    def waitFirstResponse(self, timeout):
+        start = time.time()
+        data = ""
+        while len(data) == 0:
```
pipermerriam: Maybe a cleaner way to do this

```python
start_at = time.time()
while True:
    data = self.device.read(65)
    if data:
        break
    else:
        # check timeout, sleep, etc
```

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,399 @@
...
+                blockSize = packetSize - 3 - extraHeaderSize
+            else:
+                blockSize = len(command) - offset
+            result += command[offset : offset + blockSize]
+            offset = offset + blockSize
+        while (len(result) % packetSize) != 0:
+            result += b"\x00"
+        return bytearray(result)
+
+    def unwrapResponseAPDU(self, channel, data, packetSize):
```
pipermerriam: Same with this method.  No `self` usage so it should be able to be extracted into a utility method and tested in isolation.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,399 @@
...
+            print("HID <= %s  %.2x" % (to_hex(response), sw))
+        if sw != 0x9000:
+            possibleCause = "Unknown reason"
+            if sw == 0x6982:
+                possibleCause = "Have you uninstalled the existing CA with resetCustomCA first?"
+            if sw == 0x6985:
+                possibleCause = "Condition of use not satisfied (denied by the user?)"
+            if sw == 0x6a84 or sw == 0x6a85:
+                possibleCause = "Not enough space?"
+            if sw == 0x6484:
```
pipermerriam: These should probably all be `elif` clauses.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,399 @@
...
+                dataLength = len(response) - 2
+                self.device.set_nonblocking(True)
+                break
+            result.extend(bytearray(self.device.read(65)))
+        sw = (result[swOffset] << 8) + result[swOffset + 1]
+        response = result[dataStart : dataLength + dataStart]
+        if self.debug:
+            print("HID <= %s  %.2x" % (to_hex(response), sw))
+        if sw != 0x9000:
+            possibleCause = "Unknown reason"
```
pipermerriam: Can you move this to a final `else` clause.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,399 @@
...
+                result = response
+                dataStart = 0
+                swOffset = len(response) - 2
+                dataLength = len(response) - 2
+                self.device.set_nonblocking(True)
+                break
+            result.extend(bytearray(self.device.read(65)))
+        sw = (result[swOffset] << 8) + result[swOffset + 1]
+        response = result[dataStart : dataLength + dataStart]
+        if self.debug:
```
pipermerriam: Can you drop the `debug` flag and instead use the standard python logging facilities.  We don't ship  with `print` statements.

```
class LedgerUsbDevice:
    logger = logging.getLogger('eth_account.signers.ledger.LedgerUsbDevices')

    def some_method(self):
        self.logging.debug('the message')
```

bargst:
I think the use of the hidapi is the way to go, so I spent some time cleanly implementing it with comments. Hope it looks like it could be possibly merged this way ? :)

I still have to move `test_ledger.py` to `tests/integration/ledger/...`  directory and make a README for tests usage.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+class LedgerUsbException(Exception):
+    pass
+
+
+class LedgerUsbDevice:
+    '''
+    References:
+    - https://github.com/LedgerHQ/blue-loader-python/blob/master/ledgerblue/comm.py#L56
```
dylanjw: Should the `comm` reference be taken out now that it is no longer used, or is it an important example for how the ledger works?

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,362 @@
...
+    '''
+
+    logger = logging.getLogger('eth_account.signers.ledger.LedgerUsbDevices')
+
+    def __init__(self):
+        hidDevicePath = None
+        for hidDevice in hid.enumerate(0, 0):
+            if hidDevice['vendor_id'] == 0x2c97:
+                if ('interface_number' in hidDevice and hidDevice['interface_number'] == 0) \
+                   or ('usage_page' in hidDevice and hidDevice['usage_page'] == 0xffa0):
```
dylanjw: Could these hex values be put in global constants?  `LEDGER_VENDOR_ID` and `LEDGER_USAGE_PAGE_ID`?

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,362 @@
...
+        reply_start = time.time()
+        while True:
+            packet = bytes(self.device.read(64))
+            (channel, tag, index, size, data) = unwrap_apdu(packet)
+
+            # Wait for a valid channel in replied packet
+            if not channel:
+                if reply_start + timeout < time.time():
+                    raise LedgerUsbException(f'Timeout waiting for a device' +
+                                             f'response (timeout={timeout}s)')
```
dylanjw: TIL about f-strings!

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,362 @@
...
+
+            # Check if we have received all the reply from device
+            if len(reply) > reply_min_size:
+                reply = bytes(reply[:reply_min_size])
+                break
+
+        # Status is stored at then end of the reply
+        (status,) = struct.unpack('>H', reply[-2:])
+
+        if status == 0x9000:
```
dylanjw: What does `0x9000` mean? A constant var could explain this.

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,362 @@
...
+        Sign a transaction, as in :meth:`~eth_account.account.Account.signTransaction`
+        but without specifying the private key.
+        '''
+        bip32_path = self._path_to_bytes(self.path_prefix + str(self.account_id))
+
+        unsigned_transaction = serializable_unsigned_transaction_from_dict(transaction_dict)
+        rlp_encoded_tx = rlp.encode(unsigned_transaction)
+
+        # https://github.com/LedgerHQ/blue-app-eth/blob/master/doc/ethapp.asc#sign-eth-transaction
+        apdu = bytes.fromhex('e0040000')
```
dylanjw: 'e0040000' could be saved to a constant, with a descriptive name like: `ADPU_SIGN_TX_FIRST_DATA_BLOCK`

eth_account/signers/ledger.py
```diff
@@ -0,0 +1,362 @@
...
+        apdu += bytes([len(bip32_path) + len(rlp_encoded_tx)])
+        apdu += bip32_path
+        apdu += rlp_encoded_tx
```
dylanjw: What happens if the rlp_encoded_tx is greater than the 255 byte limit?  Should this be getting chunked and sent with the next data block adpu ('e0048000')?

eth_account/signers/ledger.py
carver: Yeah, they're pretty cool. I can't wait to use them more broadly! Unfortunately f-strings weren't added until 3.6, but Web3.py (and therefore eth-account) has to support py3.5. So this will have to be switched out for `%` or `.format` style instead. :(

bargst: That explain why circleCI lint was failing ! Too bad we are not on 3.6 :)
I will update to remove all f-strings.

eth_account/signers/ledger.py
bargst: The chunk is done by the exchange() function, but I will add a test to see if it is working as expected.

bargst:
@dylanjw you where right ! Transaction larger than 255 would not be signed. Now fixed !

eth_account/signers/ledger.py
bargst: I would rather let it as is because part of the comm module are still present (device detection for ex.)

eth_account/signers/ledger.py
bargst: The sanity test has been added

bargst:
I finally found a way to make defunctSignMessage() work and sign message on the ledger !

I think the PR is now worth reviewing @pipermerriam @carver @dylanjw but do you want me to squash history before or after review ?

This PR now use hidapi python module and do not rely anymore on ledgerblue.

carver:
Squashing can happen after. I should be able to do another review tomorrow.

carver:
... and by tomorrow I mean next week. Sorry for the delay. I've been pretty heads-down on getting the next trinity release out the door.

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+)
+from eth_account.messages import (
+    defunct_hash_message,
+)
+from eth_account.signers.ledger import (
+    LedgerAccount,
+)
+
+# Not a fixture because it does not support multiple LedgerAccount
+ledger = LedgerAccount()
```
carver: What does it mean that "it does not support multiple LedgerAccount"? Maybe it should be a session fixture, like:

```py
@pytest.fixture(scope="session")
def ledger:
  return LedgerAccount()
```

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.fixture
+def acct(request):
+    return Account
```
carver: Looks like this could just be `Account` down below, and drop the `acct` fixture altogether.

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+
+def test_sign_transaction(transaction, tx_hash, acct):
+    expected_sender = ledger.address
+    signed = ledger.signTransaction(transaction)
+
+    assert type(signed.v) is int
+    assert type(signed.r) is str
+    assert type(signed.s) is str
+
+    assert signed.hash == tx_hash
```
carver: This approach looks problematic. The `tx_hash` will be different for every ledger, because it will be a different account signing it. So if I ran it on my own ledger, the test would fail. Am I missing something?

I guess one way to make it work would be to include a test setup with a (burned) mnemonic that everyone can use to initialize their ledger to the same account before the test.

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+    assert len(accounts_10) == 10
+    assert accounts_5 == accounts_10[:5]
+
+    accounts_4p0 = ledger.get_addresses(limit=4, page=0)
+    accounts_4p1 = ledger.get_addresses(limit=4, page=1)
+    accounts_4p2 = ledger.get_addresses(limit=4, page=2)
+    assert len(accounts_4p1) == 4
+    assert accounts_4p0 == accounts_10[:4]
+    assert accounts_4p1 == accounts_10[4:8]
+    assert accounts_10 == (accounts_4p0 + accounts_4p1 + accounts_4p2[:2])
```
carver: Maybe just do `accounts_12` instead of `accounts_10` above, then all three of these assertions can be dropped for: `assert accounts_12 == accounts_4p0 + accounts_4p1 + accounts_4p2`.

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+    assert accounts_1[0] == accounts_5[0]
+
+    accounts_10 = ledger.get_addresses(10)
+    assert len(accounts_10) == 10
+    assert accounts_5 == accounts_10[:5]
+
+    accounts_4p0 = ledger.get_addresses(limit=4, page=0)
+    accounts_4p1 = ledger.get_addresses(limit=4, page=1)
+    accounts_4p2 = ledger.get_addresses(limit=4, page=2)
+    assert len(accounts_4p1) == 4
```
carver: Might as well check the length of all three

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+    expected_sender = ledger.address
+    signed = ledger.signTransaction(transaction)
+
+    assert type(signed.v) is int
+    assert type(signed.r) is str
+    assert type(signed.s) is str
+
+    assert signed.hash == tx_hash
+
+    assert acct.recoverTransaction(signed.rawTransaction).lower() == expected_sender
```
carver: `ledger.address` should return the address in checksum format. Then this doesn't need to be `lower()`d in the assertion.

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+    assert accounts_10 == (accounts_4p0 + accounts_4p1 + accounts_4p2[:2])
+
+
+def test_sign_transaction(transaction, tx_hash, acct):
+    expected_sender = ledger.address
+    signed = ledger.signTransaction(transaction)
+
+    assert type(signed.v) is int
+    assert type(signed.r) is str
+    assert type(signed.s) is str
```
carver: `r` and `s` should also be returned from `signTransaciton()` as `int`, like http://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.account.Account.signTransaction

Assuming r and s are hex strings, you can convert with `eth_utils.to_int(hexstr=r)`

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
...
+    msghash = defunct_hash_message(text=message)
+
+    expected_signer = ledger.address
+
+    signed = ledger.defunctSignMessage(text=message)
+
+    assert signed.messageHash == msghash
+    assert type(signed.v) is int
+    assert type(signed.r) is str
+    assert type(signed.s) is str
```
carver: All three should be `int`, like http://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.account.Account.signHash

tests/integration/ledger/test_ledger.py
```diff
@@ -0,0 +1,116 @@
+# Require a connected Ledger ...
```
carver: Maybe even better to point to the readme, like:
```
# This test requires a connected Ledger
# For setup instructions see tests/integration/ledger/README.md
```

tests/integration/ledger/README.md
```diff
@@ -0,0 +1,13 @@
+To run the test a ledger device must be connected and the Ethereum application running on it.
+
+Install ledger requirement: 
+```sh
+# in the project root:
+pip install -e .[ledger]
+```
+
+Run the tests with:
```
carver: Probably need to set up the ledger somehow first right? From the other comments, maybe we need a shared seed for testing.

bargst:
@carver I updated following your review. I tested using the seed and it worked on both a nanos and a ledger blue using he hidapi approach !

bargst:
@carver seems my PR missed the 0.3.0 release :)

bargst:
@carver @pipermerriam do you think this PR still needs some work ?

pipermerriam:
@bargst sorry for our silence, :eyes:

pipermerriam:
@bargst thanks so much for this.  Merged in https://github.com/ethereum/eth-account/pull/34

This pr, and the new HDAccount one both fall into a category where we'll need to do an in-depth review and security audit before we can include in master and release for public use.  Our current workload suggests it'll be later this year after Devcon before we can do that so we've merged into a local feature branch until that time.

Thank you for your contribution.  It's going to be a big improvement.

bargst:
Thanks @pipermerriam for merging ! I will close this PR, feel free to ping me during review/audit :)

Issue #24, opened 2018-06-04
Implement BIP32 HD account class.


pipermerriam:
### What was wrong?

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki

BIP32 specifies Hierarchical Deterministic Wallets.  These are common ways to generate private keys.

### How can it be fixed?

Implement a new account class `HDAccount` which exposes the same API (potentially via subclassing) as [`LocalAccount`](https://github.com/ethereum/eth-account/blob/master/eth_account/local.py), with additional APIs for generating child accounts.

Test cases should include any public test vectors that are available as part of the spec.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.68 ETH (406.11 USD @ $597.22/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/eth-account/24/566).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $10,521.58 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [dilatebrave](https://gitcoin.co/profile/dilatebrave) has started work. 	
 * Q: I will do it in 1 week

darkdarkdragon:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.68 ETH (470.63 USD @ $692.1/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](http://localhost:8000/issue/ethereum/eth-account/24/755).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $0 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

pipermerriam:
[...]

PR #27, opened 2018-06-07
[WIP] Implement BIP32 HD account class

dilatebrave:
## What was wrong?

Issue #24

## How was it fixed?

- [ ] Implement a new account class HDAccount which exposes the same API (potentially via subclassing) as LocalAccount, 
- [x] Additional APIs for generating child accounts.
- [x] Test cases should include any public test vectors that are available as part of the spec.

#### Cute Animal Picture
b1178afe6c24d9a36cb5dfcfed630e14

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "${workspaceFolder}/venv/bin/python"
+}
--- a/eth_account/hdaccount.py
+++ b/eth_account/hdaccount.py
@@ -51,7 +51,7 @@ class HDAccount(object):
     # Static initializers to create from entropy
     @staticmethod
     def fromEntropy(entropy):
-        "Create a BIP32Key using supplied entropy >= MIN_ENTROPY_LEN"
+        "Create a HDAccount using supplied entropy >= MIN_ENTROPY_LEN"
         if entropy == None:
             entropy = os.urandom(MIN_ENTROPY_LEN/8) # Python doesn't have os.random()
         if not len(entropy) >= MIN_ENTROPY_LEN/8:
@@ -65,7 +65,7 @@ class HDAccount(object):
     # Normal class initializer
     def __init__(self, secret, chain, depth, index, fpr):
         """
-        Create a public or private BIP32Key using key material and chain code.
+        Create a public or private HDAccount using key material and chain code.
         secret   This is the source material to generate the keypair, either a
                  32-byte string representation of a private key, or the ECDSA
                  library object representing a public key.
@@ -117,7 +117,7 @@ class HDAccount(object):
         cK = self.PublicKey()
         return hashlib.new('ripemd160', sha256(cK).digest()).digest()[:4]
 
-    def ChildKey(self, i):
+    def ChildKey(self, i, harden=False):
         """
         Create a child key of index 'i'.
         If the most significant bit of 'i' is set, then select from the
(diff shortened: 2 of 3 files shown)


carver:
#28 has some thoughts about what the abstract class looks like. I wouldn't recommend subclassing `LocalAccount`. Just meet the interface for now. If there's any code duplication, we can pull it into utility methods.

carver:
#29 has a preview of the class that should be subclassed: `BaseAccount`

vs77bb:
@dilatebrave Are you still working here?

dilatebrave:
@vs77bb --- i'm sill working on this but got stuck at signing trans with the extended key. I'm lacking knowledge at this area so it took me sometime to read more

eth_account/hdaccount.py
```diff
@@ -0,0 +1,154 @@
...
+
+from hashlib import sha256
+from ecdsa.curves import SECP256k1
+from ecdsa.ecdsa import int_to_string, string_to_int
+
+MIN_ENTROPY_LEN = 128        # bits
+BIP32_HARDEN    = 0x80000000 # choose from hardened set of child keys
+CURVE_GEN       = ecdsa.ecdsa.generator_secp256k1
+CURVE_ORDER     = CURVE_GEN.order()
+EX_MAIN_PRIVATE = [ codecs.decode('0488ade4', 'hex') ] # Version strings for mainnet extended private keys
```
pipermerriam: Can you link to a reference for where this magic number comes from?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "${workspaceFolder}/venv/bin/python"
```
pipermerriam: This looks like something that needs to be removed and added to the `.gitignore`

eth_account/hdaccount.py
```diff
@@ -0,0 +1,154 @@
+import os
+import hmac
+import hashlib
+import struct
+import codecs
+import ecdsa
+
+from hashlib import sha256
+from ecdsa.curves import SECP256k1
```
pipermerriam: We prefer to be flexible about what implementation people choose for this sort of thing.  We have a library called [`eth-keys`](https://github.com/ethereum/eth-keys) which implements the secp256k1 operations for public/private keys and signatures that I believe is going to be our preferred way to do this.

However, I'm not entirely sure it will do everything that you need for HD keys so I'm interested in your feedback on what is missing/needed if that is the case.

vs77bb:
Hi @dilatebrave mind responding to comments back here?

eth_account/hdaccount.py
dilatebrave: These numbers come from the bip https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
I don't know how to specify the line in that file so please go there and search for the numbers >.<

.vscode/settings.json
dilatebrave: yes, that needs to be removed

eth_account/hdaccount.py
dilatebrave: i didn't know about that library. i will try it out ^^

eth_account/hdaccount.py
carver: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format <- that link is sufficient, and should go in a source comment

dilatebrave:
@vs77bb --- thanks for notifying me
@pipermerriam --- my apologies for the slow reply

pipermerriam:
@dilatebrave no apologies necessary.  Thank you for taking this on.

PR #29, opened 2018-06-07
Abstract BaseAccount for new signers to inherit 

### LINKED ISSUE #28: Create new abstract base class for LocalAccount
With the incoming work on #25 (in PR #26) -- it's time to have an abstract class defining the interface for a [`LocalAccount`](https://github.com/ethereum/eth-account/blob/master/eth_account/local.py), plus some docs about which methods must be implemented. Specifically:

- `address` property is required
- `signHash()` is required
- `signTransaction()` is required
- `__hash__()` and `__eq__()` are required - so care should be taken that the account cannot be mutated to use a different private key.
- `privateKey` property is optional, since some sources refuse to reveal it (like a hardware wallet)
- `encrypt()` is optional, since some accounts cannot reveal the private key

(Also, I just noticed that the docs for `encrypt` in `LocalAccount` are wrong. See [the definition in `Account` instead](https://github.com/ethereum/eth-account/blob/afd7a7a228974e2f1c2bb3198f7c6ae3906a8e7a/eth_account/account.py#L131-L166)

carver:
## What was wrong?

Fixes #28 

## How was it fixed?

New abstract class and docs. Moved `LocalAccount` to `signers` module.

Seeking feedback: maybe rename `*Account` to `*Signer`?

#### Cute Animal Picture

### DIFF
--- a/eth_account/signers/base.py
+++ b/eth_account/signers/base.py
@@ -1,0 +1,48 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
+
+
+class BaseAccount(ABC):
+    '''
+    A collection of convenience methods to sign transactions and message hashes.
+
+    :var str address: the checksummed public address for this account
+
+    .. code-block:: python
+
+        >>> my_account.address
+        "0xF0109fC8DF283027b6285cc889F5aA624EaC1F55"
+
+    Subclasses of :cls:`BaseAccount` should implement `__hash__` and `__eq__`,
+    where two accounts are the same if they can sign for the same address.
+    '''
+
+    @property
+    @abstractmethod
+    def address(self):
+        '''
+        The checksummed public address for this account.
+        '''
+        pass
+
+    @abstractmethod
+    def signHash(self, message_hash):
+        '''
+        Sign the hash of a message, as in :meth:`~eth_account.account.Account.signHash`
+        but without specifying the private key.
+
+        :var bytes message_hash: 32 byte hash of the message to sign
+        '''
+        pass
+
+    @abstractmethod
+    def signTransaction(self, transaction_dict):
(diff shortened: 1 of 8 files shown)


eth_account/signers/base.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @abstractmethod
+    def signTransaction(self, transaction_dict):
+        '''
+        Sign a transaction, as in :meth:`~eth_account.account.Account.signTransaction`
+        but without specifying the private key.
+
+        :var dict transaction_dict: transaction with all fields specified
+        '''
+        pass
```
dylanjw: y no abstract methods for `__eq__` and `__hash__` ?  Would it be good to force these to be implemented, or are there cases where they can't be?

carver: I guess that since `address` is required, the others should always be possible. Will add.

eth_account/signers/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __eq__(self, other):
+        '''
+        Equality test between two accounts.
+
+        Two accounts are considered the same if they are exactly the same type,
+        and can sign for the same address.
+        '''
+        return type(self) == type(other) and self.address == other.address
```
pipermerriam: I'm not sure the `type` check is correct here.  There's an argument to say that if the address is equal then the accounts are equal?

carver: I had the some thought, and am open to it. Here's why I didn't:

A hardware key account and a local private key have different capabilities, so it's conservative to treat them as unequal. I'm not really sure when those two would be tested for equality, and what people would expect, so I figured `False` was better in my ignorance than `True`.

Issue #22, opened 2018-05-16
Accept eth_keys.datatypes.PrivateKey anywhere that private keys are accepted


carver:
In addition to accepting hex strings and bytes, eth_account should accept private keys generated by eth_keys.

stefanmendoza:
@carver I'll try tackling this!

stefanmendoza:
@carver this can be closed 👍

PR #30, opened 2018-06-30
Support eth_keys.datatypes.PrivateKey in params

stefanmendoza:
## What was wrong?

Fixes issue #22 

Most private_key arguments only except `hex str`, `bytes`, or `int`. This allows the[`eth_keys.datatypes.PrivateKey`](https://github.com/ethereum/eth-keys/blob/v0.2.0-beta.3/eth_keys/datatypes.py#L219-L238) class to be provided as well.

## How was it fixed?

Updated methods in `eth_account.account.Account` to support `eth_keys.datatypes.PrivateKey` as a provided private key.

#### Cute Animal Picture


stefanmendoza:
Starting to work through this. Just a first pass right now. Going to try thinking through if there's a good way to handle this without just checking the type of the param. Also need to finish updating the tests for all the methods that were updated (only added one `Account.privateKeyToAccount` so far).

stefanmendoza:
Did some profiling. [Last commit on master](https://github.com/ethereum/eth-account/commit/6133e1ed08d8f36e401b350dbb65b8cef52205ca):
```
>>> cProfile.run("Account.privateKeyToAccount('0xb25c7db31feed9122727bf0939dc769a96564b2de4c4726d035b36ecf1e5b364')")
         1539 function calls (1284 primitive calls) in 0.006 seconds
```

This PR (SHA: bc798a94883f3e6e3a49c176aa7c7887dd30b39e):
```
>>> cProfile.run("Account.privateKeyToAccount('0xb25c7db31feed9122727bf0939dc769a96564b2de4c4726d035b36ecf1e5b364')")
         1543 function calls (1288 primitive calls) in 0.005 seconds

>>> cProfile.run("Account.privateKeyToAccount(pkey)")
         1421 function calls (1166 primitive calls) in 0.005 seconds
```

stefanmendoza:
@carver open to some feedback on this 👍 

Mainly wondering about some of the test setup / naming. Also, is there any performance testing I need to do (getting stats by profiling the changed methods, etc.)?

stefanmendoza:
@carver @pipermerriam open to some feedback on this 👍 

Mainly wondering about some of the test setup / naming and how to document it in the params list for the methods that were changed.

Also, is there any performance testing I need to do (getting stats by profiling the changed methods, etc.)?

tests/core/test_accounts.py
```diff
@@ -82,13 +95,20 @@ def test_eth_account_create_variation(acct):
...
-    key = b'a' * 32
-    acct1 = acct.privateKeyToAccount(key)
-    acct2 = acct.privateKeyToAccount(key)
+def test_eth_account_equality(acct, PRIVATE_BYTES):
+    acct1 = acct.privateKeyToAccount(PRIVATE_BYTES)
+    acct2 = acct.privateKeyToAccount(PRIVATE_BYTES)
     assert acct1 == acct2
 
 
+def test_eth_account_privateKeyToAccount_fromPrivateKey(acct, PRIVATE_BYTES, PRIVATE_KEY):
```
pipermerriam: I think that our preferred way to do this would be as a parametrized fixture, or a parametrized test.

```
@pytest.fixture(params=[bytes, eth_keys.PrivateKey])
def private_key(request):
    return request.param(PRIVATE_KEY_BYTES)

# or

@pytest.mark.parametrize(
    'private_key',
    (
        PRIVATE_KEY_AS_BYTES,
        PRIVATE_KEY_AS_OBJ,
    ),
)
def test_the_thing(private_key):
    ...
```

```

Comment (stefanmendoza) [reply tests/core/test_accounts.py]: Updated here: 3938255

Let me know if the use of those constants is okay or if we'd rather just define them in the parameterize list.

**Comment by stefanmendoza** [comment]
@pipermerriam addressed your comment. Also, should we be squashing our commits before merging?

**Comment by stefanmendoza** [comment]
@pipermerriam bumping my last question 👍

**Comment by pipermerriam** [comment]
No need to squash.

**Comment by pipermerriam** [comment]
@carver since this is `eth-account` and it's more sensitive than other code, can you review too

File: tests/core/test_accounts.py [hunk 2]
```diff
@@ -362,24 +401,56 @@ def test_eth_account_recover_transaction_from_eth_test(acct, transaction):
...
+    [
+        (
+            web3js_key(),
+            web3js_password(),
+            to_bytes(hexstr=web3js_key()),
+        ),
+        (
+            web3js_private_key(web3js_key()),
+            web3js_password(),
+            web3js_private_key(web3js_key()),
```
carver: Same here, should by `HexBytes` type.

tests/core/test_accounts.py
```diff
@@ -362,24 +401,56 @@ def test_eth_account_recover_transaction_from_eth_test(acct, transaction):
...
```
carver: [The return type of `acct.decrypt` should always be `HexBytes`](http://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.account.Account.decrypt) (this appears to be expecting a `PrivateKey`).

tests/core/test_accounts.py
```diff
@@ -51,21 +54,32 @@
...
+PRIVATE_KEY_AS_BYTES = b'unicorns' * 4
+PRIVATE_KEY_AS_BYTES_ALT = b'rainbows' * 4
+PRIVATE_KEY_AS_OBJ = keys.PrivateKey(PRIVATE_KEY_AS_BYTES)
+PRIVATE_KEY_AS_OBJ_ALT = keys.PrivateKey(PRIVATE_KEY_AS_BYTES_ALT)
 
 
-@pytest.fixture
-def PRIVATE_BYTES_ALT(PRIVATE_BYTES):
-    return b'rainbows' * 4
+@pytest.fixture(params=[PRIVATE_KEY_AS_BYTES, PRIVATE_KEY_AS_OBJ])
```
carver: Extra credit: test hexstr and int keys as well.

```
In [6]: w3.toHex(b'unicorns' * 4)
Out[6]: '0x756e69636f726e73756e69636f726e73756e69636f726e73756e69636f726e73'

In [7]: w3.toHex(b'rainbows' * 4)
Out[7]: '0x7261696e626f77737261696e626f77737261696e626f77737261696e626f7773'
```

eth_account/account.py
```diff
@@ -192,7 +196,11 @@ def privateKeyToAccount(self, private_key):
             # They correspond to the same-named methods in Account.*
             # but without the private key argument
         '''
-        key_bytes = HexBytes(private_key)
+        if isinstance(private_key, self._keys.PrivateKey):
+            key_bytes = private_key.to_bytes()
+        else:
+            key_bytes = HexBytes(private_key)
```
carver: I suppose this could follow the pattern you used elsewhere, of using the `PrivateKey` directly as `key_obj` if `private_key` is already the right type.

tests/core/test_accounts.py
stefanmendoza: Also:
```
(venv) [06:38:54 PM] @stefanmendoza ➜  eth-account git:(issue-22) ✗ python
Python 3.6.5 (default, Jun 17 2018, 12:13:06) 
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from eth_keys import keys
>>> from hexbytes import HexBytes
>>> pkey = keys.PrivateKey(HexBytes('0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'))
>>> type(pkey)
<class 'eth_keys.datatypes.PrivateKey'>
>>> type(pkey.to_bytes())
<class 'hexbytes.main.HexBytes'>
>>> pkey == pkey.to_bytes()
True
```

tests/core/test_accounts.py
stefanmendoza: Along the same lines as the comment above:
```
>>> decrypted_key
b'L\x08\x83\xa6\x91\x02\x93}b1G\x1b]\xbbb\x04\xfeQ)ap\x82y*\xe4h\xd0\x1a?6#\x18'
>>> expected_decrypted_key
'0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
>>> type(decrypted_key)
<class 'bytes'>
>>> type(expected_decrypted_key)
<class 'eth_keys.datatypes.PrivateKey'>
>>> keys.PrivateKey(decrypted_key)
'0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
>>> decrypted_key == expected_decrypted_key
True
>>> type(decrypted_key) == type(expected_decrypted_key)
False
```

tests/core/test_accounts.py
carver:
```
>>> pkey == pkey.to_bytes()
True
```

I had a feeling that ^ might be true, but didn't remember without looking it up.

Just for clarity in the test, I think we should assert the type in the test, since it's made explicit in the docs. Also, let's use an expected result that is explicitly bytes (rather than a `PrivateKey` that happens to `==` bytes).

stefanmendoza: Do we want to force the type to be `HexBytes` here?
```
return HexBytes(decode_keyfile_json(keyfile, password_bytes))
```
Or would we rather just leave it as
```
return decode_keyfile_json(keyfile, password_bytes)
```
and flex on the `bytes` type?

carver: Let's do the first: force it to `HexBytes`. Obviously this type bug predates the PR. Thanks for helping ID and fix it! :)

stefanmendoza: Fixed here: https://github.com/ethereum/eth-account/pull/30/commits/b501da3a3ba80fb9e62ce58b20575ca91f2548e7#diff-1de692d823115945c2232e1aa3066811

tests/core/test_accounts.py
stefanmendoza: Fixed here: https://github.com/ethereum/eth-account/pull/30/commits/b501da3a3ba80fb9e62ce58b20575ca91f2548e7#diff-1de692d823115945c2232e1aa3066811

eth_account/account.py
stefanmendoza: I added a new private method to add some reusability since I was repeating the pattern of generating a `eth_keys.datatypes.PrivateKey`. Open to suggestion on the naming / docs. This will also allow us to preserve the `ValueError` method as it would be difficult to follow the same pattern in this snippet while preserving the ValueError check (without having to have some repetitive type-checking code).

https://github.com/ethereum/eth-account/pull/30/commits/b501da3a3ba80fb9e62ce58b20575ca91f2548e7#diff-a9c6ca5d3c0ddc6c626bcee1142503c9R448

tests/core/test_accounts.py
stefanmendoza: I'll look into this tomorrow! Good call.

carver: I'll hold off on merging since you wanted to take a look today. Let me know if you change your mind.

eth_account/account.py
carver: Looks great, nothing to add!

tests/core/test_accounts.py
stefanmendoza: Thanks for waiting! Expanded the tests here:
https://github.com/ethereum/eth-account/pull/30/commits/335e992fa68f51d6489978620280317a3945ac3d

I ended up starting from the private key as a hexstr as the base value and creating the int, bytes, and `PrivateKey` from that. The steps I took to get the two static values for the primary and alt test values were essentially:
```
> from eth_account import Account
> acct1 = Account.create()
> address1 = acct1.address
> pkey1 = acct1.privateKey # take the hexstr from within the HexBytes(...) return value
```

stefanmendoza:
@carver responded to your comment about expanding the fixtures to include `int` and `hexstr` 👍

tests/core/test_accounts.py
```diff
@@ -54,18 +54,25 @@
 ]
 
 
-PRIVATE_KEY_AS_BYTES = b'unicorns' * 4
-PRIVATE_KEY_AS_BYTES_ALT = b'rainbows' * 4
```
carver: Oh no, why did you kill the unicorns and rainbows? :cry:

I may revive them later... For now, I'll just merge as is.

stefanmendoza: My main issue was generating a `hexstr` from those :/

stefanmendoza: I think we can generate a `PrivateKey` from those and get the `hexstr` from that 🤔

carver: The hexstr for them was in my comment asking about testing int and hexstrs.

carver: So the block could be:

```py
PRIVATE_KEY_AS_BYTES = b'unicorns' * 4
PRIVATE_KEY_AS_HEXSTR = '0x756e69636f726e73756e69636f726e73756e69636f726e73756e69636f726e73'
PRIVATE_KEY_AS_INT = 0x756e69636f726e73756e69636f726e73756e69636f726e73756e69636f726e73
PRIVATE_KEY_AS_OBJ = keys.PrivateKey(PRIVATE_KEY_AS_BYTES)
ACCT_ADDRESS = '0xa79F6f349C853F9Ea0B29636779ae3Cb4E3BA729'
```

stefanmendoza: Yeah, I meant the account address, not the `hexstr`. Generated a `PrivateKey` from the bytes and got address from the account. 👍 Added them back!

stefanmendoza: I'll remove the casting and represent the int as the primitive.

carver:
Hooray!

PR #38, opened 2018-10-11
Allow KDF to be provided for private key encryption

### LINKED ISSUE #3: Allow specifying KDF function for `web3.eth.account.encrypt` API
Migrated from: https://github.com/ethereum/web3.py/issues/583

### What was wrong?

`eth-keyfile` supports passing in a `kdf` parameter to specify the key derivation function, but this api is not exposed via the `web3.eth.account.encrypt` API.

### How can it be fixed?

Add an optional keyword parameter `kdf` to allow specifying what KDF you want.

Extra options:

Allow setting the default KDF using both environment variables and via a property on the `web3.eth.account` module.

```python
web3.eth.account.defaultKDF = 'scrypt'
```

### LINKED ISSUE #4: Use scrypt as default
>    web3.py uses pbkdf2(dklen:32, c:1000000, prf: hmac-sha256).
    these are pretty low defaults
    should def at least use scrypt as a baseline
    hmac-sha256 is much easier to bruteforce, by like 3-4 orders of magnitude
    the parameters don't matter i think the KDF func/lib should be able to accommodate but probably best to make them match the geth defaults
    which are more secure. for some comparison an iteration of the default SCRYPT parameters in geth takes 200-400ms but an iteration of the HMAC KDF takes less than 1ms, on GPUs SCRYPT with r=8 destroys performance because of the high memory bandwidth > requirement, so even on super high end GPUs you can barely get 20-30h/s where as with the HMAC KDF you can get 100k-1m h/s a second.

stefanmendoza:
## What was wrong?

Fixes #3 
Fixes #4 

## How was it fixed?

- Updated the `Account#encrypt` and `LocalAccount#encrypt` methods to take an optional 'kdf' parameter to be used when generating the encrypted keyfile (Issue #3)
- Switched to using 'scrypt' as the default KDF for keyfile encryption instead of the 'pbkdf2' default from eth-keyfile (Issue #4). If the environment variable `ETH_ACCOUNT_KDF` is set, that will be used for the default KDF if none is provided when calling `Account#encrypt` or `LocalAccount#encrypt`

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -51,7 +51,6 @@ from eth_account.signers.local import (
 
 
 class Account(object):
-
     '''
     This is the primary entry point for working with Ethereum private keys.
 
@@ -134,7 +133,6 @@ class Account(object):
         password_bytes = text_if_str(to_bytes, password)
         return HexBytes(decode_keyfile_json(keyfile, password_bytes))
 
-    # TODO - Update documentation to reflect new KDF param
     @staticmethod
     def encrypt(private_key, password, kdf=None):
         '''
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -3,10 +3,6 @@ from collections import (
 )
 import json
 import os
-import sys
-from types import (
-    ModuleType,
-)
 
 from cytoolz import (
     dissoc,
@@ -54,25 +50,12 @@ from eth_account.signers.local import (
 )
 
 
-class _AccountModule(ModuleType):
+class Account(object):
 
     '''
     TODO - Documentation
     '''
-    _defaultKDF = os.getenv('ETH_ACCOUNT_KDF', 'scrypt')
-
-    @property
-    def defaultKDF(self):
-        '''
-        TODO - Documentation
-        '''
-        return type(self)._defaultKDF
-
-
-sys.modules[__name__].__class__ = _AccountModule
-
-
-class Account(object):
+    default_kdf = os.getenv('ETH_ACCOUNT_KDF', 'scrypt')
 
     '''
     This is the primary entry point for working with Ethereum private keys.
@@ -151,7 +134,7 @@ class Account(object):
 
     # TODO - Update documentation to reflect new KDF param
     @staticmethod
-    def encrypt(private_key, password, kdf=getattr(sys.modules[__module__], 'defaultKDF')):  # noqa: E501,F821
+    def encrypt(private_key, password, kdf=None):
         '''
         Creates a dictionary with an encrypted version of your private key.
         To import this keyfile into Ethereum clients like geth and parity:
@@ -192,6 +175,9 @@ class Account(object):
             key_bytes = private_key.to_bytes()
         else:
             key_bytes = HexBytes(private_key)
+
+        if kdf is None:
+            kdf = __class__.default_kdf  # noqa: F821
 
         password_bytes = text_if_str(to_bytes, password)
         assert len(key_bytes) == 32
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -53,16 +53,18 @@ from eth_account.signers.local import (
 class Account(object):
 
     '''
-    TODO - Documentation
-    '''
-    default_kdf = os.getenv('ETH_ACCOUNT_KDF', 'scrypt')
-
-    '''
     This is the primary entry point for working with Ethereum private keys.
 
     It does **not** require a connection to an Ethereum node.
     '''
     _keys = keys
+
+    default_kdf = os.getenv('ETH_ACCOUNT_KDF', 'scrypt')
+    '''
+    The default key deriviation function (KDF) to use when encrypting a private key. If the
+    environment variable :envvar:`ETH_ACCOUNT_KDF` is set, it's value will be used as the default.
+    Otherwise, 'scrypt' will be used as the default.
+    '''
 
     @combomethod
     def create(self, extra_entropy=''):
@@ -144,6 +146,7 @@ class Account(object):
         :param private_key: The raw private key
         :type private_key: hex str, bytes, int or :class:`eth_keys.datatypes.PrivateKey`
         :param str password: The password which you will need to unlock the account in your client
+        :param str kdf: The key derivation function to use when encrypting your private key
         :returns: The data to use in your encrypted file
         :rtype: dict
 
@@ -155,18 +158,27 @@ class Account(object):
                 getpass.getpass()
             )
 
-            {'address': '5ce9454909639d2d17a3f753ce7d93fa0b9ab12e',
-             'crypto': {'cipher': 'aes-128-ctr',
-              'cipherparams': {'iv': '78f214584844e0b241b433d7c3bb8d5f'},
-              'ciphertext': 'd6dbb56e4f54ba6db2e8dc14df17cb7352fdce03681dd3f90ce4b6c1d5af2c4f',
-              'kdf': 'pbkdf2',
-              'kdfparams': {'c': 1000000,
-               'dklen': 32,
-               'prf': 'hmac-sha256',
-               'salt': '45cf943b4de2c05c2c440ef96af914a2'},
-              'mac': 'f5e1af09df5ded25c96fcf075ada313fb6f79735a914adc8cb02e8ddee7813c3'},
-             'id': 'b812f3f9-78cc-462a-9e89-74418aa27cb0',
-             'version': 3}
+            {
+                'address': '5ce9454909639d2d17a3f753ce7d93fa0b9ab12e',
+                'crypto': {
+                    'cipher': 'aes-128-ctr',
+                    'cipherparams': {
+                        'iv': '0b7845a5c3597d3d378bde9b7c7319b7'
+                    },
+                    'ciphertext': 'a494f1feb3c854e99c1ff01e6aaa17d43c0752009073503b908457dc8de5d2a5',  # noqa: E501
+                    'kdf': 'scrypt',
+                    'kdfparams': {
+                        'dklen': 32,
+                        'n': 262144,
+                        'p': 8,
+                        'r': 1,
+                        'salt': '13c4a48123affaa29189e9097726c698'
+                    },
+                    'mac': 'f4cfb027eb0af9bd7a320b4374a3fa7bef02cfbafe0ec5d1fd7ad129401de0b1'
+                },
+                'id': 'a60e0578-0e5b-4a75-b991-d55ec6451a6f',
+                'version': 3
+            }
 
              >>> with open('my-keyfile', 'w') as f:
                  f.write(json.dumps(encrypted))
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -3,6 +3,10 @@ from collections import (
 )
 import json
 import os
+import sys
+from types import (
+    ModuleType,
+)
 
 from cytoolz import (
     dissoc,
@@ -50,7 +54,26 @@ from eth_account.signers.local import (
 )
 
 
+class _AccountModule(ModuleType):
+
+    '''
+    TODO - Documentation
+    '''
+    _defaultKDF = os.getenv('ETH_ACCOUNT_KDF', 'scrypt')
+
+    @property
+    def defaultKDF(self):
+        '''
+        TODO - Documentation
+        '''
+        return type(self)._defaultKDF
+
+
+sys.modules[__name__].__class__ = _AccountModule
+
+
 class Account(object):
+
     '''
     This is the primary entry point for working with Ethereum private keys.
 
@@ -126,8 +149,9 @@ class Account(object):
         password_bytes = text_if_str(to_bytes, password)
         return HexBytes(decode_keyfile_json(keyfile, password_bytes))
 
+    # TODO - Update documentation to reflect new KDF param
     @staticmethod
-    def encrypt(private_key, password):
+    def encrypt(private_key, password, kdf=getattr(sys.modules[__module__], 'defaultKDF')):  # noqa: E501,F821
         '''
         Creates a dictionary with an encrypted version of your private key.
         To import this keyfile into Ethereum clients like geth and parity:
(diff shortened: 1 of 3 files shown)


tests/core/test_accounts.py
```diff
@@ -409,54 +427,81 @@ def test_eth_account_recover_transaction_from_eth_test(acct, transaction):
...
             web3js_private_key(web3js_key()).to_bytes(),
+            'scrypt'
         ),
     ],
-    ids=['hex_str', 'eth_keys.datatypes.PrivateKey']
+    ids=['hex_str', 'hex_str_provided_kdf', 'eth_keys.datatypes.PrivateKey']
 )
-def test_eth_account_encrypt(acct, private_key, password, expected_decrypted_key):
-    encrypted = acct.encrypt(private_key, password)
+def test_eth_account_encrypt(acct, private_key, password, kdf, expected_decrypted_key, expected_kdf):  # noqa: E501
```
stefanmendoza: Are we fine with silencing this flake warning here and below or is there some pattern established for how we format method signatures that are too long per the linting?

tests/core/test_accounts.py
```diff
@@ -97,6 +99,22 @@ def acct(request):
...
+
+
+def test_eth_account_default_kdf_from_env():
+    import importlib
+    os.environ['ETH_ACCOUNT_KDF'] = 'pbkdf2'
+    importlib.reload(eth_account.account)
+    assert eth_account.account.defaultKDF == 'pbkdf2'
+
+    # This is done to ensure we use the default (scrypt)
+    # KDF for subsequent tests
```
stefanmendoza: Need to dig into this... I'm confused why this doesn't break subsequent tests that don't provide a KDF to `#encrypt` as I would assume the default for those would be `pdkdf2` unless (a) this module reloading / env updating is locally scoped or (b) the `_defaultKDF = os.getenv('ETH_ACCOUNT_KDF' , 'scrypt')` is lazily evaluated.

stefanmendoza:
@pipermerriam / @carver  I'm ready to have the implementation / tests looked over for this. All that's left is updating the doc / some cleanup after my questions are answered. 👍

eth_account/signers/local.py
```diff
@@ -26,7 +26,7 @@ class LocalAccount(BaseAccount):
     def __init__(self, key, account):
         '''
         :param eth_keys.PrivateKey key: to prefill in private key execution
-        :param web3.account.Account account: the key-unaware management API
+        :param ~eth_account.account.Account account: the key-unaware management API
```
stefanmendoza: Updated to point to the correct class.

tests/core/test_accounts.py
pipermerriam: Generally we break the line rather than silencing, dropping each argument on it's own line like this: https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#function-signatures

tests/core/test_accounts.py
```diff
@@ -97,6 +99,22 @@ def acct(request):
     return Account
 
 
+def test_eth_account_default_kdf():
+    assert eth_account.account.defaultKDF == 'scrypt'
+
+
+def test_eth_account_default_kdf_from_env():
+    import importlib
+    os.environ['ETH_ACCOUNT_KDF'] = 'pbkdf2'
```
pipermerriam: For this type of thing you should use the `monkeypatch` fixture provided by pytest

https://docs.pytest.org/en/latest/monkeypatch.html

Allows you to modify environment variables in a way that they are always cleaned up.

tests/core/test_accounts.py
pipermerriam: Yeah, that is odd and yes, I would expect tests to break.  You might throw a debugger statement into another test to see why it isn't.

eth_account/account.py
```diff
@@ -50,7 +57,30 @@
...
+        TODO - Documentation
+        '''
+        return type(self)._defaultKDF
+
+    @defaultKDF.setter
+    def defaultKDF(self, kdf):
+        type(self)._defaultKDF = kdf
+
+
+sys.modules[__name__].__class__ = _AccountModule
```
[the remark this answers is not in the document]
stefanmendoza: I can switch it. I initally wanted to put it on one of the Account classes, but the issue statement:
>Allow setting the default KDF using both environment variables and via a property on the web3.eth.account module.
>
> `web3.eth.account.defaultKDF = 'scrypt'`

Made it seem like you wanted it on `eth_account.account` (the module) and not `eth_account.account.Account` (the class).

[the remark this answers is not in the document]
pipermerriam: You correctly interpreted my original comment/spec (which I didn't recall until you just pointed it out), but seeing it in action, I'm thinking that class level default with option for global default via environment variable is the right way to go.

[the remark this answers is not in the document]
carver: Note that in the web3 context, `w3.eth.account` is an instance of the `Account` class, so changing the default works smoothly as `w3.eth.account.defaultKDF = 'scrypt'`.

tests/core/test_accounts.py
carver: Weirdly the `setenv` method disappeared from recent docs. Here's an old copy with the reference: https://docs.pytest.org/en/3.0.0/monkeypatch.html

tests/core/test_accounts.py
```diff
@@ -409,54 +427,81 @@ def test_eth_account_recover_transaction_from_eth_test(acct, transaction):
...
         ),
     ],
-    ids=['hex_str', 'eth_keys.datatypes.PrivateKey']
+    ids=['hex_str', 'hex_str_provided_kdf', 'eth_keys.datatypes.PrivateKey']
 )
-def test_eth_account_prepared_encrypt(acct, private_key, password, expected_decrypted_key):
+def test_eth_account_prepared_encrypt(acct, private_key, password, kdf, expected_decrypted_key, expected_kdf):  # noqa: E501
     account = acct.privateKeyToAccount(private_key)
-    encrypted = account.encrypt(password)
+    encrypted = account.encrypt(password) if kdf is None else account.encrypt(password, kdf)
```
carver: For readability, I usually recommend something like the following instead of the inline ternary operator:
```py
if kdf is None:
  encrypted = account.encrypt(password)
else:
  encrypted = account.encrypt(password, kdf)
```

The closest thing I can find in the style guide is https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#one-thing-per-line ... it's probably worth checking if there is broad consensus on that and adding it to the style guide.

An exception might be something like: `num_threads = 0 if cpu_count is None else cpu_count` -- even then I might slightly prefer the other syntax, but not say anything on a PR.

stefanmendoza: Yeah, I thought the ternary operator would be preferred, but for the test above this one ([see here](https://github.com/ethereum/eth-account/pull/38/files/d8ee9af6fd0467d2e25ea028c2ab6e2145d2dcfe#diff-1de692d823115945c2232e1aa3066811R457)) it caused me to break the line length limit so I used the regular conditional blocks. I agree that'd probably be something good to add to the style guide. 👍  I'll switch it to be consistent since the above would need to use the regular conditional blocks unless we ignored flake warnings (which I'm guessing we'd rather not do if it can be avoided). Thanks!

tests/core/test_accounts.py
stefanmendoza: Fixed here
https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-1de692d823115945c2232e1aa3066811L456
and here
https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-1de692d823115945c2232e1aa3066811L498

tests/core/test_accounts.py
stefanmendoza: Still need to dig into this to make sure we're still not in that weird state, but the test structure has changed:
https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-1de692d823115945c2232e1aa3066811R101

eth_account/account.py
```diff
@@ -126,8 +156,9 @@ def decrypt(keyfile_json, password):
         password_bytes = text_if_str(to_bytes, password)
         return HexBytes(decode_keyfile_json(keyfile, password_bytes))
 
+    # TODO - Update documentation to reflect new KDF param
     @staticmethod
-    def encrypt(private_key, password):
+    def encrypt(private_key, password, kdf=getattr(sys.modules[__module__], 'defaultKDF')):  # noqa: E501,F821
```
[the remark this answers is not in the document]
stefanmendoza: Updated here:
https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-a9c6ca5d3c0ddc6c626bcee1142503c9R137

eth_account/account.py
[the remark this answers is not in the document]
stefanmendoza: So I'm still a little unclear about the ask here...

I have it setup so it's a class variable that gets instantiated with `os.getenv('ETH_ACCOUNT_KDF', 'scrypt')`. To set the env var after the class had already been imported, I had to reload the module and use the new Account reference to verify this functionality ([see here](https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-1de692d823115945c2232e1aa3066811R109)). Based on the docs [here](https://docs.python.org/3/library/importlib.html#importlib.reload):
>Other references to the old objects (such as names external to the module) are not rebound to refer to the new objects and must be updated in each namespace where they occur if that is desired.

I think this is the proper way to do it as the
```
from eth_account import (
    Account,
)
```
at the top of the file would be the original reference, where the class object has `default_kdf = 'scrypt'`. Still kind of new to Python, so please let me know if my understanding of this is off.

Another question I had @carver, based on your comment, is if we want this value to be settable vs. something that's just instantiated with a default and that default can be overridden by just providing a KDF to the `encrypt` method. I'm thinking that making it mutable just adds more complexity since the only place it's used is the `encrypt` method and that already allows you to provide whatever KDF you want.

Commit: https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-a9c6ca5d3c0ddc6c626bcee1142503c9

tests/core/test_accounts.py
stefanmendoza: Updated the tests, but refer to this comment:
https://github.com/ethereum/eth-account/pull/38#discussion_r224943961

tests/core/test_accounts.py
stefanmendoza: Fixed: https://github.com/ethereum/eth-account/pull/38/commits/8aec325d1b02c045143b0106e6a807524c04c6b6#diff-1de692d823115945c2232e1aa3066811R509

stefanmendoza:
@carver updated the docs: https://github.com/ethereum/eth-account/pull/38/commits/2a12f8242dd4107c4b7daa5bf9b851e681477edb

One last question I had (mentioned [here](https://github.com/ethereum/eth-account/pull/38#discussion_r224943961)) was whether or not we do want the `default_kdf` to be mutable like your example from Web3? If so, then I think we're good. 👍

eth_account/account.py
```diff
@@ -126,8 +134,9 @@ def decrypt(keyfile_json, password):
         password_bytes = text_if_str(to_bytes, password)
         return HexBytes(decode_keyfile_json(keyfile, password_bytes))
 
+    # TODO - Update documentation to reflect new KDF param
```
carver: Yup mutable attribute for default is okay. Also can you delete this line?

carver:
This is getting into the territory where it would be nice to squash commits (with a few tiny fixups), but I'll just merge this time. :+1: Thanks!

stefanmendoza:
@carver yeah, my normal pattern is keep the commits that I put in comments until the final approval. Next time once I get the good to go I'll squash if it seems like there's a lot of commit. Thanks! 👍

eth_account/account.py
```diff
@@ -169,9 +186,12 @@ def encrypt(private_key, password):
         else:
             key_bytes = HexBytes(private_key)
 
+        if kdf is None:
+            kdf = __class__.default_kdf  # noqa: F821
```
pipermerriam: I never knew that `__class__` was available like this within a method bodies!

stefanmendoza: Curious (since I'm new-ish to Python and still learning 😁), what made you think it wouldn't be? I used it since it was a `staticmethod` and referencing the class using `__class__` seemed reasonable as we're not calling a method on the `cls`, which is how I've typically seen methods annotated with `@classmethod` used.

Searching around, I find contradictory things though:
>Static methods are great for utility functions, which perform a task in isolation. **They don’t need to (and cannot) access class data.** They should be completely self-contained, and only work with data passed in as arguments.

but also
>The `@staticmethod` decorator is similar to `@classmethod` in that it can be called from an uninstantiated class object, although in this case there is no `cls` parameter passed to its method. 
>
>Since no `self` object is passed either, that means we also don't have access to any instance data, and thus this method can not be called on an instantiated object either.
> 
>These types of methods aren't typically meant to create/instantiate objects, **but they may contain some type of logic pertaining to the class itself**, like a helper or utility method

I think this works because the method is still defined on the class and the `default_kdf` variable is a global / class variable, but now I'm wondering if this is an anti-pattern or not. 🤔

carver: Yeah, I missed this. My preferred approach would be to change the method to a `@classmethod` and use `cls.default_kdf` here. It's a little less magical.

pipermerriam: @stefanmendoza I'd probably have done `type(self)` but I realize now that this was a `staticmethod`.  @carver suggestion to change this to a class method seems like the right approach.

stefanmendoza: I can put out a quick follow-up PR 👍

PR #41, opened 2018-10-17
Switch Account#encrypt to @classmethod

stefanmendoza:
## What was wrong?

Switch to a `@classmethod` for `Account#encrypt` instead of calling `__class__` from inside a `@staticmethod`.

Reference: https://github.com/ethereum/eth-account/pull/38#discussion_r225709792

## How was it fixed?

Switch to `@staticmethod`, call `cls.default_kdf`.

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -133,8 +133,8 @@ class Account(object):
         password_bytes = text_if_str(to_bytes, password)
         return HexBytes(decode_keyfile_json(keyfile, password_bytes))
 
-    @staticmethod
-    def encrypt(private_key, password, kdf=None):
+    @classmethod
+    def encrypt(cls, private_key, password, kdf=None):
         '''
         Creates a dictionary with an encrypted version of your private key.
         To import this keyfile into Ethereum clients like geth and parity:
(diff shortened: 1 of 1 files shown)


eth_account/account.py
```diff
@@ -187,7 +187,7 @@ def encrypt(private_key, password, kdf=None):
             key_bytes = HexBytes(private_key)
 
         if kdf is None:
-            kdf = __class__.default_kdf  # noqa: F821
+            kdf = cls.default_kdf  # noqa: F821
```
carver:
```suggestion
            kdf = cls.default_kdf
```

Should be able to drop the flake8 annotation now

stefanmendoza:
@carver ah, forgot this wasn't on 1.x yet. w00t 🎉

eth_account/account.py
stefanmendoza: Removed 👍 

https://github.com/ethereum/eth-account/pull/41/commits/690bbdd0c59c5bc567fd03cbc0b06a0b677424b5#diff-a9c6ca5d3c0ddc6c626bcee1142503c9L190

Issue #42, opened 2018-10-18
Move internal module to _utils


carver:
## What was wrong?

Our most recent pattern for tools that are not guaranteed to be API stable is to put them in a `_utils` module. `internal` is used for that purpose in eth-account, but it would be nice to use the new pattern.

## How can it be fixed?

Rename `internal` -> `_utils`, and update all the appropriate imports.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 45.0 DAI (45.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/eth-account/42/1543).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/eth-account/42/1543).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $49,234.38 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 week, 6 days from now.
Please review their action plans below:


**1) [mayurva](https://gitcoin.co/profile/mayurva) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/eth-account/42/1543?mutate_worker_action=approve&worker=mayurva) | [reject worker](https://gitcoin.co/issue/ethereum/eth-account/42/1543?mutate_worker_action=reject&worker=mayurva))_.**

So basically planning to
1. Rename the internal module to _util
2. Update all the references to it
3. Ensure the tests still pass

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/eth-account/42/1543).

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__Work for 45.0 DAI (45.0 USD @ $1.0/DAI) has been submitted by__:

 1. [@anshumanv](https://gitcoin.co/profile/anshumanv)



@ceresstation please take a look at the submitted work:
[...]

PR #45, opened 2018-10-19
Renamed the internal module and all it's references to point to _utils

omrsmeh:
## What was wrong?

Issue #42 

## How was it fixed?

Renamed the internal module and all it's references to point to _utils


pipermerriam:
fixed with #46

Issue #9, opened 2018-03-27
Set up docs to run doctests in python examples


carver:
See quirk in #7 

Otherwise, it should be similar to getting the doctests running in web3.py.

nickventon:
Hey @carver, mind assigning this one to me? I’m interested in working on it.

pacrob:
Thanks, @nickventon, but looks like this was handled in #94, should have been closed earlier.

PR #49, opened 2018-11-01
Run doctests with tox

AndreMiras:
## What was wrong?

Currently the CI doesn't run doctests.
However keep in mind that now we have 5 tests failing.
If you're interested with the approach, I can also fix them in this PR or in another one.

Issue #9 

## How was it fixed?
Enabling doctests via `tox.ini` and `pytest.ini`

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -106,21 +106,21 @@ class Account(object):
         .. code-block:: python
 
             >>> encrypted = {
-             'address': '5ce9454909639d2d17a3f753ce7d93fa0b9ab12e',
-             'crypto': {'cipher': 'aes-128-ctr',
-              'cipherparams': {'iv': '78f214584844e0b241b433d7c3bb8d5f'},
-              'ciphertext': 'd6dbb56e4f54ba6db2e8dc14df17cb7352fdce03681dd3f90ce4b6c1d5af2c4f',
-              'kdf': 'pbkdf2',
-              'kdfparams': {'c': 1000000,
-               'dklen': 32,
-               'prf': 'hmac-sha256',
-               'salt': '45cf943b4de2c05c2c440ef96af914a2'},
-              'mac': 'f5e1af09df5ded25c96fcf075ada313fb6f79735a914adc8cb02e8ddee7813c3'},
-             'id': 'b812f3f9-78cc-462a-9e89-74418aa27cb0',
-             'version': 3}
+            ... 'address': '5ce9454909639d2d17a3f753ce7d93fa0b9ab12e',
+            ... 'crypto': {'cipher': 'aes-128-ctr',
+            ...  'cipherparams': {'iv': '78f214584844e0b241b433d7c3bb8d5f'},
+            ...  'ciphertext': 'd6dbb56e4f54ba6db2e8dc14df17cb7352fdce03681dd3f90ce4b6c1d5af2c4f',
+            ...  'kdf': 'pbkdf2',
+            ...  'kdfparams': {'c': 1000000,
+            ...   'dklen': 32,
+            ...   'prf': 'hmac-sha256',
+            ...   'salt': '45cf943b4de2c05c2c440ef96af914a2'},
+            ...  'mac': 'f5e1af09df5ded25c96fcf075ada313fb6f79735a914adc8cb02e8ddee7813c3'},
+            ... 'id': 'b812f3f9-78cc-462a-9e89-74418aa27cb0',
+            ... 'version': 3}
 
             >>> import getpass
-            >>> Account.decrypt(encrypted, getpass.getpass())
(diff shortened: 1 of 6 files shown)


pipermerriam:
Yes please, enabling doctest is something we've been adding to all of our repos and it should be in place here as well.

carver:
Yup, this is great, thanks!

> If you're interested with the approach, I can also fix them in this PR or in another one.

Please fix in this PR. As a rule we never merge a failing CI run.

FWIW, it's important that the readability of the docs not suffer in order to get the tests passing.

AndreMiras:
Great, then I may pick this up over the weekend if I get a chance.
Yes I agree the documentation should still be as readable :+1:

fubuloubu:
Seems like this can be closed?

carver:
> Seems like this can be closed?

I don't think this change was done yet. We are already test appropriately-marked snippets in the `docs/` folder. But this one would test the docstring'd method definitions, which is also reasonable to do.

AndreMiras:
Hi guys, I will eventually resume my contributions starting from this one as I have more time :tada: 
So I've rebased this PR, fixed the tests I could and skipped some others. Currently 9 are skipped, but I think it's a good start for now and we may reduce this number in the future.
I can't wait to contribute more in the future :smile:

eth_account/account.py
```diff
@@ -106,21 +106,21 @@ def decrypt(keyfile_json, password):
...
+            ...  'cipherparams': {'iv': '78f214584844e0b241b433d7c3bb8d5f'},
+            ...  'ciphertext': 'd6dbb56e4f54ba6db2e8dc14df17cb7352fdce03681dd3f90ce4b6c1d5af2c4f',
+            ...  'kdf': 'pbkdf2',
+            ...  'kdfparams': {'c': 1000000,
+            ...   'dklen': 32,
+            ...   'prf': 'hmac-sha256',
+            ...   'salt': '45cf943b4de2c05c2c440ef96af914a2'},
+            ...  'mac': 'f5e1af09df5ded25c96fcf075ada313fb6f79735a914adc8cb02e8ddee7813c3'},
+            ... 'id': 'b812f3f9-78cc-462a-9e89-74418aa27cb0',
+            ... 'version': 3}
```
AndreMiras: Enabling doctests helped in detecting this "syntax error"

eth_account/account.py
```diff
@@ -106,21 +106,21 @@ def decrypt(keyfile_json, password):
...
 
             >>> import getpass
-            >>> Account.decrypt(encrypted, getpass.getpass())
+            >>> Account.decrypt(encrypted, getpass.getpass()) # doctest: +SKIP
```
AndreMiras: I had to disable because obviously `getpass()` would ask for terminal input.
We could also give an example with an hardcoded value for the `password` that always produces an expected output.

carver: Yeah, maybe worth doing both: demonstrating good practices but skipping the test (as you already did), and adding a version that can be doctest-ed.

carver:
> Currently 9 are skipped

Maybe add an issue for fixing & re-enabling these tests?

eth_account/messages.py
```diff
@@ -187,13 +187,13 @@ def encode_defunct(
         SignableMessage(version=b'E', header=b'thereum Signed Message:\n6', body=b'I\xe2\x99\xa5SF')
 
         # these four also produce the same hash:
-        >>> encode_defunct(w3.toBytes(text=message_text))
+        >>> encode_defunct(w3.toBytes(text=message_text)) # doctest: +SKIP
```
carver: Just curious what the failure was here (I'm fine to push it to another issue)

AndreMiras:
```
NameError: name 'w3' is not defined
```

eth_account/account.py
AndreMiras: Yes that could be it. Actually I initially wanted to find the password used to generate this output, but haven't found it so far.
I've created a task for adding a doctest that works https://github.com/ethereum/eth-account/issues/81

AndreMiras:
Thanks for the review and approval @carver 

> I'll wait a bit for you to add that non-getpass() version. If you want to skip that for now, just let me know, and I'll merge it in.

I'm on holiday without much access to dev laptop, so I'd prefer we merge it. However I've created a [followup task](https://github.com/ethereum/eth-account/issues/81) that I'll be happy to pick up later.

eth_account/messages.py
carver: Hah, ok. Well that's not an instance method so it should be straightforward to switch `w3` to `Web3` and just add a `from web3 import Web3` up top. Just a note for when you/someone does the follow-up work.

Issue #48, opened 2018-10-31
Account.create() pbkdf2 iterations/work-factor access


AndreMiras:
[Account.encrypt()](https://github.com/ethereum/eth-account/blob/8cb7d4e/eth_account/account.py#L137) doesn't provide means to set the `iterations` (work-factor) parameter of the [eth_keyfile.keyfile.create_keyfile_json()](https://github.com/ethereum/eth-keyfile/blob/v0.5.1/eth_keyfile/keyfile.py#L37) method.
It would be useful if that was somewhat possible without monkey patching the call.
My use case is I'm running Python-based "Ethereum clients" on smart-phones (e.g. [EtherollApp](https://github.com/AndreMiras/EtherollApp), [PyWallet](https://github.com/AndreMiras/PyWallet)) and it has less resources so account creation with default pbkdf2 params may take ages.
I was initially using [pyethereum](https://github.com/ethereum/pyethereum) where it was somewhat possible overriding `PBKDF2_CONSTANTS['c']`, but I'm in the process of migrating it since it's deprecated.

They way around I'm currently using with `eth-account` is to monkey patch `create_keyfile_json()` call.
```python
import eth_account
from eth_keyfile import create_keyfile_json

account = eth_account.Account.create()
eth_account.account.create_keyfile_json = partial(
    create_keyfile_json, iterations=1)
encrypted = eth_account.Account.encrypt(account.privateKey, "password")
```
I know setting `iterations=1` is terribly wrong but it's for demo-purpose. It could be useful for speeding up unit testing too.

Anyway, if the use-case of giving access to this param makes sense, I could make a pull request.

pipermerriam:
I'm :+1: on figuring out a way to allow these parameters to be explicitly set.

stefanmendoza:
@AndreMiras I put out a PR for this:
https://github.com/ethereum/eth-account/pull/52

Feel free to check it out and let me know what you think! 👍

AndreMiras:
I eventually updated to recent `eth-account` and I confirm this is working OK, I can take out my monkey patch, thanks :smile:

PR #52, opened 2019-01-01
Allow iterations (work factor) to be provided to Account#encrypt

stefanmendoza:
## What was wrong?

* Resolves #48
* **(EDIT: Moved to #53)**`pytest-xdist 1.25.0 has requirement pytest>=3.6.0, but you'll have pytest 3.3.2 which is incompatible.` ([1.24.1](https://github.com/pytest-dev/pytest-xdist/blob/v1.24.1/setup.py#L3) vs. [1.25.0](https://github.com/pytest-dev/pytest-xdist/blob/v1.25.0/setup.py#L3))

## How was it fixed?

* Expose `iterations` as an optional param to `Account#encrypt`
* **(EDIT: Moved to #53)** Uplift `pytest` to `>=3.6.0` and fix the resulting breaking changes in tests accordingly
* Tests added for providing the work factor to Account.encrypt

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #53, opened 2019-01-05
Upgrade pytest to >= 3.6.0

stefanmendoza:
## What was wrong?

Prerequisite for Issue #48 

There is no restriction on pytest-dist version, which introduces the
following error:

pytest-xdist 1.25.0 has requirement pytest>=3.6.0, but you'll have pytest 3.3.2 which is incompatible.

When upgrading to pytest >= 3.6.0, fixtures can no longer be called
directly as functions and thus the broken tests resulting from the
version bump of pytest had to be fixed.

## How was it fixed?

Bump pytest to `>= 3.6.0` and fix tests accordingly.

#### Cute Animal Picture

### DIFF
--- a/tests/core/test_accounts.py
+++ b/tests/core/test_accounts.py
@@ -91,21 +91,6 @@ def PRIVATE_KEY_ALT(request):
 
 
 @pytest.fixture
-def web3js_key():
-    return '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
-
-
-@pytest.fixture
-def web3js_private_key(web3js_key):
-    return keys.PrivateKey(HexBytes(web3js_key))
-
-
-@pytest.fixture
-def web3js_password():
-    return 'test!'
-
-
-@pytest.fixture
 def acct(request):
     return Account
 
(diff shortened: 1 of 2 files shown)


stefanmendoza:
cc @davesque / @pipermerriam / @carver

tests/core/test_accounts.py
```diff
@@ -434,31 +419,44 @@ def test_eth_account_recover_transaction_from_eth_test(acct, transaction):
...
-@pytest.mark.parametrize(
-    'private_key, password, kdf, expected_decrypted_key, expected_kdf',
-    [
+def encrypt_mark_params():
+    """
+    Params for testing Account#encrypt. Due to not being able to provie fixtures to
+    pytest.mark.parameterize, we opt for creating the params in a non-fixture method
+    here instead of providing fixtures for the public key, password, and private key.
+    """
+    public_key = '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
```
carver: This is a private key, the only difference between `web3js_key` and `web3js_private_key` was that one is a hex string and the other is a `keys.PrivateKey`. So the variable name should be updated.

tests/core/test_accounts.py
```diff
@@ -90,21 +90,6 @@ def PRIVATE_KEY_ALT(request):
     return request.param
 
 
-@pytest.fixture
-def web3js_key():
```
carver: A long time ago this used to be used as a fixture, eg~ https://github.com/ethereum/eth-account/commit/d18f1454e0b9c48dd8c7be0086933e7327fb0e7b#diff-1de692d823115945c2232e1aa3066811R346

Then at some point, the method was converted from being used as a fixture, to being used as a simple method, but the change slipped in without removing the fixture decorator. Looks like here: https://github.com/ethereum/eth-account/commit/bc798a94883f3e6e3a49c176aa7c7887dd30b39e

Perhaps a simpler solution is to remove the `fixture` decorators here, and the next two methods. (Looks like this method could use a comment to clarify that this key is also the private key) We just want to test that account works when being passed a hex string or a native private key. :+1:

stefanmendoza: Cool cool, I can add a follow-up PR for this. 👍

stefanmendoza: Will actually update on #52 since that one is still open. Thanks!

stefanmendoza: See: https://github.com/ethereum/eth-account/pull/52#issuecomment-453379730

PR #56, opened 2019-02-20
Implement signing data with intended validator as part of EIP 191

Bhargavasomu:
## What was wrong?

`EIP 191` needs to be implemented as part of https://github.com/ethereum/web3.py/issues/1241

## How was it fixed?

* Personal Sign was already implemented.
* Support was added for `signature version 0x00` (Signing `Data with intended validator`) by making `defunct_hash_message` take the `signature_version` and the `version_specific_data` as arguments. Then the `signature_wrapper` function was modified to add support to these `signature versions`.

#### Cute Animal Picture

### DIFF
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -18,6 +18,10 @@ from eth_account._utils.transactions import (
 CHAIN_ID_OFFSET = 35
 V_OFFSET = 27
 
+# signature versions
+PERSONAL_SIGN_VERSION = b'E'  # Hex value 0x45
+INTENDED_VALIDATOR_SIGN_VERSION = b'\x00'  # Hex value 0x00
+
 
 def sign_transaction_dict(eth_key, transaction_dict):
     # generate RLP-serializable transaction, with defaults filled
@@ -50,11 +54,11 @@ def signature_wrapper(message, signature_version, version_specific_data):
             type(signature_version))
         )
 
-    if signature_version == b'E':
+    if signature_version == PERSONAL_SIGN_VERSION:
         preamble = b'\x19Ethereum Signed Message:\n'
         size = str(len(message)).encode('utf-8')
         return preamble + size + message
-    elif signature_version == b'\x00':
+    elif signature_version == INTENDED_VALIDATOR_SIGN_VERSION:
         wallet_address = to_bytes(hexstr=version_specific_data)
         if len(wallet_address) != 20:
             raise TypeError("Invalid Wallet Address: {}".format(version_specific_data))
@@ -63,7 +67,10 @@ def signature_wrapper(message, signature_version, version_specific_data):
     else:
         raise NotImplementedError(
             "Currently supported signature versions are: {0}, {1}. ".
-            format('0x' + b'\x00'.hex(), '0x' + b'E'.hex()) +
+            format(
+                '0x' + INTENDED_VALIDATOR_SIGN_VERSION.hex(),
+                '0x' + PERSONAL_SIGN_VERSION.hex()
+            ) +
             "But received signature version {}".format('0x' + signature_version.hex())
         )
 
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -1,4 +1,5 @@
 from cytoolz import (
+    curry,
     pipe,
 )
 from eth_utils import (
@@ -40,14 +41,25 @@ def sign_transaction_dict(eth_key, transaction_dict):
 
 
 # watch here for updates to signature format: https://github.com/ethereum/EIPs/issues/191
-def signature_wrapper(message, version=b'E'):
-    assert isinstance(message, bytes)
-    if version == b'E':
+@curry
+def signature_wrapper(message, signature_version, version_specific_data):
+    if not isinstance(message, bytes):
+        raise TypeError("Message is not of the type {}, expected bytes".format(type(message)))
+
+    if signature_version == b'E':
         preamble = b'\x19Ethereum Signed Message:\n'
         size = str(len(message)).encode('utf-8')
         return preamble + size + message
+    elif signature_version == b'\x00':
+        wallet_address = to_bytes(hexstr=version_specific_data)
+        if len(wallet_address) != 20:
+            raise TypeError("Invalid Wallet Address: {}".format(version_specific_data))
+        wrapped_message = b'\x19' + signature_version + wallet_address + message
+        return wrapped_message
     else:
-        raise NotImplementedError("Only the 'Ethereum Signed Message' preamble is supported")
+        raise NotImplementedError(
+            "Only Personal Sign and Signing of data with intended validator is supported"
+        )
 
 
 def hash_of_signed_transaction(txn_obj):
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -44,7 +44,11 @@ def sign_transaction_dict(eth_key, transaction_dict):
 @curry
 def signature_wrapper(message, signature_version, version_specific_data):
     if not isinstance(message, bytes):
-        raise TypeError("Message is not of the type {}, expected bytes".format(type(message)))
+        raise TypeError("Message is of the type {}, expected bytes".format(type(message)))
+    if not isinstance(signature_version, bytes):
+        raise TypeError("Signature Version is of the type {}, expected bytes".format(
+            type(signature_version))
+        )
 
     if signature_version == b'E':
         preamble = b'\x19Ethereum Signed Message:\n'
@@ -58,7 +62,9 @@ def signature_wrapper(message, signature_version, version_specific_data):
         return wrapped_message
     else:
         raise NotImplementedError(
-            "Only EIP 191 type signature wrapping is currently supported"
+            "Currently supported signature versions are: {0}, {1}. ".
+            format('0x' + b'\x00'.hex(), '0x' + b'E'.hex()) +
+            "But received signature version {}".format('0x' + signature_version.hex())
         )
 
 
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -58,7 +58,7 @@ def signature_wrapper(message, signature_version, version_specific_data):
         return wrapped_message
     else:
         raise NotImplementedError(
-            "Only Personal Sign and Signing of data with intended validator is supported"
+            "Only EIP 191 type signature wrapping is currently supported"
         )
 
 
--- a/eth_account/messages.py
+++ b/eth_account/messages.py
@@ -15,8 +15,8 @@ from eth_account._utils.signing import (
 
 
 def defunct_hash_message(
-        *,
         primitive=None,
+        *,
         hexstr=None,
         text=None,
         signature_version=b'E',
@@ -24,21 +24,20 @@ def defunct_hash_message(
     '''
     Convert the provided message into a message hash, to be signed.
     This provides the same prefix and hashing approach as
-    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`. The message will by default
-    be prepended with text defined in EIP-191 as
-    version 'E': ``b'\\x19Ethereum Signed Message:\\n'``
+    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`.
+    Currently you can only specify the ``signature_version`` as following.
+    version ``0x45`` (version ``E``): ``b'\\x19Ethereum Signed Message:\\n'``
     concatenated with the number of bytes in the message.
-    You can also specify the ``signature_version`` as
-        1) version ``0``: Sign data with intended validator (EIP 191)
-           Here the version_specific_data would be a hexstr which is the 20 bytes account address
-        2) version ``1``: Sign the structured data (EIP 712)
-           Here the version_specific_data would be <WIP(Needs to be updated)>
+    NOTE: This is the defualt version followed, if the signature_version is not specified.
+    version ``0x00`` (version ``0``): Sign data with intended validator (EIP 191).
+    Here the version_specific_data would be a hexstr which is the 20 bytes account address
+    of the intended validator.
 
-    Awkwardly, the number of bytes in the message is encoded in decimal ascii. So
-    if the message is 'abcde', then the length is encoded as the ascii
+    For version ``0x45`` (version ``E``), Awkwardly, the number of bytes in the message is
+    encoded in decimal ascii. So if the message is 'abcde', then the length is encoded as the ascii
     character '5'. This is one of the reasons that this message format is not preferred.
     There is ambiguity when the message '00' is encoded, for example.
-    Only use this method if you must have compatibility with
+    Only use this method with version ``E`` if you must have compatibility with
     :meth:`w3.eth.sign() <web3.eth.Eth.sign>`.
 
     Supply exactly one of the three arguments:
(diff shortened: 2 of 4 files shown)


Bhargavasomu:
@pipermerriam could you please review? Thankyou.

Bhargavasomu:
@kclowes I have a few doubts regarding this signing procedure. This PR aims to support the new signing mechanism as per `EIP 191`. But my major doubts are the following
* During the whole `lifecycle of the message` which one signs, where is it being verified? (Could you please link me the code if we have it in `web3.py`)
* Further even if it verified somewhere, where is the message being extracted properly? (Like from b'\x19Ethereum Signed Message:\n'<some message> where are we extracting the <some message> part and showing it to the user?) Because I hoped that this would be happening in `web3.py`, but I couldn't find it anywhere in my searches.

Could you please give me clarity on the above questions, so that I could do better testing with this `signing mechanism`? Thankyou.

/cc @pipermerriam

Bhargavasomu:
@pipermerriam @kclowes I believe that the verification of the signature is done through `recoverHash` and `recoverTransaction` methods. Further I believe that the **Exact Message** extraction from the encoded message, is done in the other side of the contract and not from `web3` part (Please correct me if wrong). That said, I think this PR tested that part. Could you please give this a review? Thankyou.

kclowes:
@Bhargavasomu sorry it has taken me so long to respond! I should have some time tomorrow and I'll definitely have time on Wednesday to review.

kclowes:
@Bhargavasomu I need a little more time to wrap my head around the PR. I hope I'll be able to get you some feedback/answers to questions tomorrow!

eth_account/messages.py
```diff
@@ -37,6 +48,8 @@ def defunct_hash_message(primitive=None, hexstr=None, text=None):
     :type primitive: bytes or int
     :param str hexstr: the message encoded as hex
     :param str text: the message as a series of unicode characters (a normal Py3 str)
+    :param bytes signature_version: a byte indicating which kind of prefix is to be added (EIP 191)
+    :param version_specific_data: the data which is needed for adding the prefix (EIP 191)
```
kclowes: I don't think the wording here is quite right. I don't think the data is needed for adding the prefix. I think it's just data. It is related to the prefix, but not needed for _adding_ the prefix. For example, in EIP 712 it's the domain separator. https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-domainseparator, and in EIP 191 it's an address like you have above.

eth_account/messages.py
```diff
@@ -14,14 +14,25 @@
...
-    defined in EIP-191 as version 'E': ``b'\\x19Ethereum Signed Message:\\n'``
+    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`. The message will by default
+    be prepended with text defined in EIP-191 as
+    version 'E': ``b'\\x19Ethereum Signed Message:\\n'``
     concatenated with the number of bytes in the message.
+    You can also specify the ``signature_version`` as
+        1) version ``0``: Sign data with intended validator (EIP 191)
+           Here the version_specific_data would be a hexstr which is the 20 bytes account address
+        2) version ``1``: Sign the structured data (EIP 712)
+           Here the version_specific_data would be <WIP(Needs to be updated)>
```
kclowes: The `version_specific_data` here is the domain separator https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-domainseparator

If you are leaving out EIP 712 from this PR, I'd prefer to remove the WIP here, please :)

eth_account/_utils/signing.py
```diff
@@ -40,14 +41,25 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+    elif signature_version == b'\x00':
+        wallet_address = to_bytes(hexstr=version_specific_data)
+        if len(wallet_address) != 20:
+            raise TypeError("Invalid Wallet Address: {}".format(version_specific_data))
+        wrapped_message = b'\x19' + signature_version + wallet_address + message
+        return wrapped_message
     else:
-        raise NotImplementedError("Only the 'Ethereum Signed Message' preamble is supported")
+        raise NotImplementedError(
+            "Only Personal Sign and Signing of data with intended validator is supported"
```
kclowes: Maybe we should say `Only EIP 191 is currently supported` here.

kclowes:
> * Based on above explanation, I don't see why `web3.py` has to be changed with respect to implementation of `EIP191` or `EIP712`. In other words, `web3.py` should automatically reflect the `EIP191` and `EIP712` standard, with the changes made only in this repository (and not in `web3.py`).

Yes, I think that's correct!

Bhargavasomu:
@kclowes done! Please review whenever you have time. Thankyou.

eth_account/messages.py
```diff
@@ -14,14 +14,23 @@
...
     This provides the same prefix and hashing approach as
-    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`. That means that the
-    message will automatically be prepended with text
-    defined in EIP-191 as version 'E': ``b'\\x19Ethereum Signed Message:\\n'``
+    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`. The message will by default
+    be prepended with text defined in EIP-191 as
+    version 'E': ``b'\\x19Ethereum Signed Message:\\n'``
     concatenated with the number of bytes in the message.
+    Currently you can only specify the ``signature_version`` as
+    version ``0``: Sign data with intended validator (EIP 191).
```
kclowes: `signature_version` here can be either version `0` or version `E`, right? We should probably update this sentence to reflect that. 

Maybe something like:
Currently, the only two supported signature versions are `E` and `0`. If version `0` is used, the `version_specific_data` is the address of the intended validator.

eth_account/messages.py
```diff
@@ -14,14 +14,23 @@
...
+    Here the version_specific_data would be a hexstr which is the 20 bytes account address
 
     Awkwardly, the number of bytes in the message is encoded in decimal ascii. So
```
kclowes: I think this paragraph would be better right after we talk about version E, since it only applies there. What do you think?

eth_account/messages.py
```diff
@@ -14,14 +14,23 @@
 )
 
 
-def defunct_hash_message(primitive=None, hexstr=None, text=None):
+def defunct_hash_message(
+        *,
```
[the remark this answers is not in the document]
pipermerriam: I didn't either when I first saw it.  It's like `*args` which is "gather all the remaining positional arguments", but instead it says, "disallow any positional arguments".  It's a way to require a function be called with **only** keyword arguments.

We often use it for functions with long signatures, or in this case, I believe it's intent is to require one of the `primative/hexstr/text` to be provided and to only allow them to be keywords.

Ideally, I'd like to allow `primative` to be positional, but brief experimentation doesn't seem to make that easy to do....

eth_account/messages.py
pipermerriam: I'd avoid referring to it as `E`, and rather to represent both as `0x00` and `0x45` to make it clear that it's their *byte* value that matters.  After representing them that way, you could make the connection that `0x45` is equivalent to the ASCII letter `E`

eth_account/messages.py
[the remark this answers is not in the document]
kclowes: Ohh, interesting. That's cool!

`primitive` can be positional without the `*`, though right?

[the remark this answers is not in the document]
Bhargavasomu: I'm adding only the `primitive` as the `keyword argument`. So now the function definition would look like this.
```
def defunct_hash_message(
        primitive=None,
        *,
        hexstr=None,
        text=None,
        signature_version=b'E',
        version_specific_data=None):
```

eth_account/messages.py
Bhargavasomu: @kclowes how about we first mention all the `versions` we are supporting, and then we talk about this case with mention of `version E` specifically. Just for having all of them at the same place.

Bhargavasomu:
@kclowes @pipermerriam could you please check if the `docs` are right now? I still haven't added the tests, but will add them in a while.

Bhargavasomu:
@kclowes I have added the new tests also. Please have a look and let me know. Thankyou!

tests/core/test_accounts.py
```diff
@@ -322,6 +322,63 @@ def test_eth_account_sign(acct, message, key, expected_bytes, expected_hash, v,
...
+        ("hello world", "string"),
+        (b'hello world', "bytes"),
+        ("68656c6c6f20776f726c64", "hexstr"),
+    )
+)
+def test_eth_account_sign_data_with_intended_validator(message, message_type, acct):
+    account = acct.create()
+
+    if message_type == "string":
+        primitive = None
```
kclowes: style nit: You could assign these to a variable like: `kwargs = {'text': message}`, and then pass `**kwargs` to `defunct_hash_message` below. Then, you'd avoid having to assign `None` to the args you weren't using.

tests/core/test_accounts.py
```diff
@@ -322,6 +322,63 @@ def test_eth_account_sign(acct, message, key, expected_bytes, expected_hash, v,
...
+        hashed_msg = defunct_hash_message(
+            primitive=primitive,
+            text=text,
+            hexstr=hexstr,
+            signature_version=b'\x00',
+            version_specific_data=account.address + "12345",
+        )
+    with pytest.raises(TypeError):
+        # Raise TypeError if the address is less than 20 bytes
+        hashed_msg = defunct_hash_message(
```
kclowes: Same thing here about assigning `hashed_msg`.

tests/core/test_accounts.py
```diff
@@ -322,6 +322,63 @@ def test_eth_account_sign(acct, message, key, expected_bytes, expected_hash, v,
...
+        text = None
+        hexstr = None
+    elif message_type == "hexstr":
+        primitive = None
+        text = None
+        hexstr = message
+
+    with pytest.raises(TypeError):
+        # Raise TypeError if the address is more than 20 bytes
+        hashed_msg = defunct_hash_message(
```
kclowes: Also not a big deal, but you don't have to assign `hashed_msg` here. You can just say `defunct_hash_message(...`.

Bhargavasomu:
@kclowes done, review please.

tests/core/test_accounts.py
```diff
@@ -322,6 +322,44 @@ def test_eth_account_sign(acct, message, key, expected_bytes, expected_hash, v,
...
+    (
+        ("hello world", "text"),
+        (b'hello world', "primitive"),
+        ("68656c6c6f20776f726c64", "hexstr"),
+    )
+)
+def test_eth_account_sign_data_with_intended_validator(message, message_type, acct):
+    account = acct.create()
+
+    kwargs = {message_type: message}
```
kclowes: 👍

eth_account/_utils/signing.py
```diff
@@ -40,14 +41,25 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+    elif signature_version == b'\x00':
+        wallet_address = to_bytes(hexstr=version_specific_data)
+        if len(wallet_address) != 20:
+            raise TypeError("Invalid Wallet Address: {}".format(version_specific_data))
+        wrapped_message = b'\x19' + signature_version + wallet_address + message
+        return wrapped_message
     else:
-        raise NotImplementedError("Only the 'Ethereum Signed Message' preamble is supported")
+        raise NotImplementedError(
+            "Only EIP 191 type signature wrapping is currently supported"
```
pipermerriam: This message should probably include the hex representation of the `signature_version` and potentially an explicit list of the hex representations of the versions that are supported.

eth_account/_utils/signing.py
```diff
@@ -40,14 +41,25 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
 # watch here for updates to signature format: https://github.com/ethereum/EIPs/issues/191
-def signature_wrapper(message, version=b'E'):
-    assert isinstance(message, bytes)
-    if version == b'E':
+@curry
+def signature_wrapper(message, signature_version, version_specific_data):
+    if not isinstance(message, bytes):
+        raise TypeError("Message is not of the type {}, expected bytes".format(type(message)))
+
+    if signature_version == b'E':
```
pipermerriam: The `b'E'` should probably be encoded as a constant with a semantically meaningful name rather than using the literal inline here.  Same with the null byte used below for the other version.

eth_account/messages.py
```diff
@@ -14,20 +14,30 @@
...
     Convert the provided message into a message hash, to be signed.
     This provides the same prefix and hashing approach as
-    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`. That means that the
-    message will automatically be prepended with text
-    defined in EIP-191 as version 'E': ``b'\\x19Ethereum Signed Message:\\n'``
+    :meth:`w3.eth.sign() <web3.eth.Eth.sign>`.
+    Currently you can only specify the ``signature_version`` as following.
+    version ``0x45`` (version ``E``): ``b'\\x19Ethereum Signed Message:\\n'``
     concatenated with the number of bytes in the message.
+    NOTE: This is the defualt version followed, if the signature_version is not specified.
```
pipermerriam: To get this to render properly in the docs this should be updated to use the following syntax.

```rst
.. note: This is the default version ...
```

tests/core/test_accounts.py
```diff
@@ -322,6 +322,44 @@ def test_eth_account_sign(acct, message, key, expected_bytes, expected_hash, v,
...
+
+    with pytest.raises(TypeError):
+        # Raise TypeError if the address is more than 20 bytes
+        defunct_hash_message(
+            **kwargs,
+            signature_version=b'\x00',
+            version_specific_data=account.address + "12345",
+        )
+    with pytest.raises(TypeError):
+        # Raise TypeError if the address is less than 20 bytes
```
pipermerriam: We like highly granular tests.  The reasoning being that it's better to not have multiple failures hiding behind a single failure.  I.E. Were both of these assertions to be failing (the method accepted both short and too-long addresses, then you wouldn't see the short-address failure until you fixed the long-address failure.

Ideally we want individual test cases for each of the *conditions* we test which suggest this test function should be split into three:

- test too-short address
- test too-long address
- test *valid* address.

Bhargavasomu:
@pipermerriam done. Please review again. Thankyou.

eth_account/messages.py
```diff
@@ -28,7 +28,7 @@ def defunct_hash_message(
     Currently you can only specify the ``signature_version`` as following.
     version ``0x45`` (version ``E``): ``b'\\x19Ethereum Signed Message:\\n'``
     concatenated with the number of bytes in the message.
-    NOTE: This is the defualt version followed, if the signature_version is not specified.
+    .. note:: This is the defualt version followed, if the signature_version is not specified.
     version ``0x00`` (version ``0``): Sign data with intended validator (EIP 191).
```
pipermerriam: @Bhargavasomu can you build the docs locally and verify this renders correctly.  I'm thinking there's some whitespace needed here but not sure exactly where and how much.

tests/core/test_accounts.py
```diff
@@ -360,6 +343,52 @@ def test_eth_account_sign_data_with_intended_validator(message, message_type, ac
...
+                version_specific_data=to_hex(address_in_bytes[:-i]),
+            )
+
+
+@pytest.mark.parametrize(
+    'message, message_type',
+    (
+        ("hello world", "text"),
+        (b'hello world', "primitive"),
+        ("68656c6c6f20776f726c64", "hexstr"),
```
pipermerriam: A good way to do this to reduce the repitition would be a parametrized fixture.

```python
@pytest.fixture(params=('text', 'primative', 'hexstr'))
def signature_kwargs(request):
    if request.params == 'text':
        return {"text": "hello world"}
    ...
```

Then in you tests you can just include `signature_kwargs` in the arguments and the test will run once for each of the fixture params.

eth_account/messages.py
Bhargavasomu: @pipermerriam I have fixed the docs, and the output is as follows.
[
Screenshot from 2019-03-11 22-23-45
](url)

Bhargavasomu:
@pipermerriam ready for another round of review. Thankyou.

Issue #35, opened 2018-09-26
Implement structured signing according to EIP712


pipermerriam:
## What was wrong?

No support for EIP712 style signed messages:  https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md

## How can it be fixed?

Implement a low level signing API for creating and verifying EIP712 style signatures with appropriate primatives for wiring up to a higher level integration into the `Account` class.

pipermerriam:
Up for discussion on what the name for this new signing method should be and what API it should expose.  Given the complexity of structured messages, I don't think it is reasonable to try and define the API up-front, but suggesting a few names for the actual method.

- `signStructured`
- `signEIP712`
- `signStructuredData`

pipermerriam:
cross linking to EIP191: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md which has a different use case (and should be implemented independently if we see forward progress on adoption)

carver:
As long as 712 is in [`Draft` status](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#eip-work-flow), the new method will need a big honking warning in the docs saying that it the hash is subject to change without a major version bump, and that it shouldn't be used in prod.

I think we're free to use `snake_case` here. Maybe `sign_typed_data`?

If the method takes in data (in whatever format) and returns the 32-byte hash, then it will be easy to use with these other eth-account methods:

- https://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.account.Account.signHash
- https://eth-account.readthedocs.io/en/latest/eth_account.html#eth_account.account.Account.recoverHash

Bhargavasomu:
@pipermerriam is this still up for me to work on? Further IIUC, then we are just adding another type of `signature mechanism` apart from the existing one. So, was this signature introduced for a specific use case (or) is it just another plugin kind of thing that the users could use?

pipermerriam:
Yes, you are still :+1: to run with this.  I'm going to delegate to @kclowes to lead figuring out details of the implementation.  My intuition is that it'll be 1-2 new signature APIs, one for the `EIP191` with `0x00` prefix, and one for the `EIP191 + EIP712` structured signing.  These APIs probably exist *both* on the `Web3` namespace somewhere, delegating to the underlying ethereum node for the signing, as well as on the `eth-account` API for locally managed private keys, potentially with a corresponding middleware that does the proper interception of messages to perform local signing.

[...]

PR #57, opened 2019-03-18
Implement EIP 712

Bhargavasomu:
## What was wrong?

Fixes Issue #35

## How was it fixed?

[Example.js](https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.js) of the [EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) was taken as reference.

### TODOS
1. ~Supporting the scenario where the `data` could be `arrays` (Including error handling etc.)~
2. ~Validation of the `JSON Structured data`. (With the use of `Regex`)~
3. ~Appropriate Error Handling. (Especially in case where the data is not of the **specified** corresponding type)~
4. ~Add `tests` for the **failing or Exception rising cases**~
5. ~Add `tests` for the `Array Data` and its corresponding `Errors` or `Exceptions`.~

#### Cute Animal Picture

### DIFF
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -4,14 +4,14 @@ from cytoolz import (
     curry,
     pipe,
 )
+from eth_abi import (
+    encode_abi,
+)
 from eth_utils import (
     keccak,
     to_bytes,
     to_int,
     to_text,
-)
-from eth_abi import (
-    encode_abi,
 )
 
 from eth_account._utils.transactions import (
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -9,7 +9,7 @@ from eth_utils import (
 )
 
 from eth_account._utils.structured_data import (
-    hashStruct,
+    hash_struct,
 )
 from eth_account._utils.transactions import (
     ChainAwareUnsignedTransaction,
@@ -73,8 +73,10 @@ def signature_wrapper(message, signature_version, version_specific_data):
         return wrapped_message
     elif signature_version == STRUCTURED_DATA_SIGN_VERSION:
         message_string = to_text(primitive=message)
-        domainSeparator = hashStruct(message_string, is_domain_separator=True)
-        wrapped_message = b'\x19' + signature_version + domainSeparator + hashStruct(message_string)
+        domainSeparator = hash_struct(message_string, is_domain_separator=True)
+        wrapped_message = (
+            b'\x19' + signature_version + domainSeparator + hash_struct(message_string)
+        )
         return wrapped_message
     else:
         raise NotImplementedError(
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -8,8 +8,10 @@ from eth_utils import (
     to_text,
 )
 
-from eth_account._utils.structured_data import (
-    hash_struct,
+from eth_account._utils.structured_data.hashing import (
+    hash_domain,
+    hash_message,
+    load_and_validate_structured_message,
 )
 from eth_account._utils.transactions import (
     ChainAwareUnsignedTransaction,
@@ -73,9 +75,10 @@ def signature_wrapper(message, signature_version, version_specific_data):
         return wrapped_message
     elif signature_version == STRUCTURED_DATA_SIGN_VERSION:
         message_string = to_text(primitive=message)
-        domainSeparator = hash_struct(message_string, is_domain_separator=True)
+        structured_data = load_and_validate_structured_message(message_string)
+        domainSeparator = hash_domain(structured_data)
         wrapped_message = (
-            b'\x19' + signature_version + domainSeparator + hash_struct(message_string)
+            b'\x19' + signature_version + domainSeparator + hash_message(structured_data)
         )
         return wrapped_message
     else:
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -91,6 +91,10 @@ def dict_to_type_name_converter(field):
 
 
 def encodeType(primaryType, types):
+    """
+    The type of a struct is encoded as name ‖ "(" ‖ member₁ ‖ "," ‖ member₂ ‖ "," ‖ … ‖ memberₙ ")"
+    where each member is written as type ‖ " " ‖ name.
+    """
     # Getting the dependencies and sorting them alphabetically as per EIP712
     deps = dependencies(primaryType, types)
     deps_without_primary_type = list(filter(lambda x: x != primaryType, deps))
@@ -98,8 +102,8 @@ def encodeType(primaryType, types):
 
     result = ''.join(
         [
-            dep + "(" + ','.join(map(dict_to_type_name_converter, types[dep])) + ")"
-            for dep in sorted_deps
+            struct_name + "(" + ','.join(map(dict_to_type_name_converter, types[struct_name])) + ")"
+            for struct_name in sorted_deps
         ]
     )
     return result
@@ -123,7 +127,7 @@ def encodeData(primaryType, types, data):
         if field["type"] == "string":
             if not isinstance(value, str):
                 raise TypeError(
-                    "Value of `{0}` ({2}) of field `{1}` is of the type `{3}`, but expected "
+                    "Value of `{0}` ({2}) in the struct `{1}` is of the type `{3}`, but expected "
                     "string value".format(
                         field["name"],
                         primaryType,
@@ -138,7 +142,7 @@ def encodeData(primaryType, types, data):
         elif field["type"] == "bytes":
             if not isinstance(value, bytes):
                 raise TypeError(
-                    "Value of `{0}` ({2}) of field `{1}` is of the type `{3}`, but expected "
+                    "Value of `{0}` ({2}) in the struct `{1}` is of the type `{3}`, but expected "
                     "bytes value".format(
                         field["name"],
                         primaryType,
@@ -163,8 +167,8 @@ def encodeData(primaryType, types, data):
             try:
                 is_encodable(field["type"], value)
             except:
-                raise AttributeError(
-                    "Received Invalid type `{0}` of field `{1}`".format(
+                raise TypeError(
+                    "Received Invalid type `{0}` in the struct `{1}`".format(
                         field["type"],
                         primaryType,
                     )
@@ -177,7 +181,7 @@ def encodeData(primaryType, types, data):
                 encValues.append(value)
             else:
                 raise TypeError(
-                    "Value of `{0}` ({2}) of field `{1}` is of the type `{3}`, but expected "
+                    "Value of `{0}` ({2}) in the struct `{1}` is of the type `{3}`, but expected "
                     "{4} value".format(
                         field["name"],
                         primaryType,
@@ -207,18 +211,18 @@ def validate_structured_data(structured_data):
                 )
 
 
-def hashStruct(structured_json_string_data, for_domain=False):
+def hashStruct(structured_json_string_data, is_domain_separator=False):
     """
     The structured_json_string_data is expected to have the ``types`` attribute and
     the ``primaryType``, ``message``, ``domain`` attribute.
-    The ``for_domain`` variable is used to calculate the ``hashStruct`` as part of the
-    ``domainSeparator`` calculation.
+    The ``is_domain_separator`` variable is used to calculate the ``hashStruct`` as
+    part of the ``domainSeparator`` calculation.
     """
     structured_data = json.loads(structured_json_string_data)
     validate_structured_data(structured_data)
 
     types = structured_data["types"]
-    if for_domain:
+    if is_domain_separator:
         primaryType = "EIP712Domain"
         data = structured_data["domain"]
     else:
@@ -227,7 +231,9 @@ def hashStruct(structured_json_string_data, for_domain=False):
     return keccak(encodeData(primaryType, types, data))
 
 
-# watch here for updates to signature format: https://github.com/ethereum/EIPs/issues/191
+# watch here for updates to signature format:
+# https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md
+# https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md
 @curry
 def signature_wrapper(message, signature_version, version_specific_data):
     if not isinstance(message, bytes):
@@ -249,8 +255,7 @@ def signature_wrapper(message, signature_version, version_specific_data):
         return wrapped_message
     elif signature_version == STRUCTURED_DATA_SIGN_VERSION:
         message_string = to_text(primitive=message)
-        # Here the version_specific_data is the EIP712Domain JSON string (includes type also)
-        domainSeparator = hashStruct(message_string, for_domain=True)
+        domainSeparator = hashStruct(message_string, is_domain_separator=True)
         wrapped_message = b'\x19' + signature_version + domainSeparator + hashStruct(message_string)
         return wrapped_message
     else:
--- a/eth_account/_utils/signing.py
+++ b/eth_account/_utils/signing.py
@@ -1,21 +1,16 @@
-import json
-import re
-
 from cytoolz import (
     curry,
     pipe,
 )
-from eth_abi import (
-    encode_abi,
-    is_encodable,
-)
 from eth_utils import (
-    keccak,
     to_bytes,
     to_int,
     to_text,
 )
 
+from eth_account._utils.structured_data import (
+    hashStruct,
+)
 from eth_account._utils.transactions import (
     ChainAwareUnsignedTransaction,
     UnsignedTransaction,
@@ -31,10 +26,6 @@ V_OFFSET = 27
 PERSONAL_SIGN_VERSION = b'E'  # Hex value 0x45
 INTENDED_VALIDATOR_SIGN_VERSION = b'\x00'  # Hex value 0x00
 STRUCTURED_DATA_SIGN_VERSION = b'\x01'  # Hex value 0x01
-
-# Regexes
-IDENTIFIER_REGEX = r"^[a-zA-Z_$][a-zA-Z_$0-9]*$"
-TYPE_REGEX = r"^[a-zA-Z_$][a-zA-Z_$0-9]*(\[([1-9]\d*)*\])*$"
 
 
 def sign_transaction_dict(eth_key, transaction_dict):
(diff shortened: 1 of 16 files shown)


Bhargavasomu:
@kclowes @pipermerriam there is still stuff to be done and this is just a `WIP`, but could you please give me a **preliminary review**? Thankyou.

fubuloubu:
@Bhargavasomu I'm working on adding EIP 712 support in a Plasma Cash client I am writting in Python here: https://github.com/zatoichi-labs/plasma-cash-vyper/tree/signed-structs

I can try and implement using your updates to see how the EIP 712 compatibility works in a practical use case.

eth_account/_utils/signing.py
```diff
@@ -44,6 +58,175 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+                    "Invalid Identifier `{}` in `{}`".format(field["name"], field_type)
+                )
+            # Check that field["type"] matches with TYPE_REGEX
+            if not re.match(TYPE_REGEX, field["type"]):
+                raise AttributeError(
+                    "Invalid Type `{}` in `{}`".format(field["type"], field_type)
+                )
+
+
+def hashStruct(structured_json_string_data, for_domain=False):
```
kclowes: What do you think about changing `for_domain` to `is_domain_separator`?

eth_account/_utils/signing.py
```diff
@@ -44,6 +58,175 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+    if for_domain:
+        primaryType = "EIP712Domain"
+        data = structured_data["domain"]
+    else:
+        primaryType = structured_data["primaryType"]
+        data = structured_data["message"]
+    return keccak(encodeData(primaryType, types, data))
+
+
 # watch here for updates to signature format: https://github.com/ethereum/EIPs/issues/191
```
kclowes: This comment is outdated. I think we can get rid of it. Or change it to EIP712 :)

tests/core/test_EIP712.py
```diff
@@ -0,0 +1,437 @@
...
+)
+def test_type_regex(type, valid):
+    if valid:
+        assert re.match(TYPE_REGEX, type) is not None
+    else:
+        assert re.match(TYPE_REGEX, type) is None
+
+
+def test_structured_data_invalid_identifier_filtered_by_regex():
+    invalid_structured_data_string = '''{
```
kclowes: I don't really have any good ideas about how to do this right at the moment, but it seems like there should be a way to be able to merge these invalid data structures with `structured_valid_data_json_string`. So instead of re-writing the whole string out, we could do something like: 
```
structured_valid_data_json_string.merge(
  {"Person": [ "name": "hello wallet", "type": "address" }] }
)
```

setup.py
```diff
@@ -46,6 +46,7 @@
     include_package_data=True,
     install_requires=[
         "attrdict>=2.0.0,<3",
+        "eth-abi",
```
kclowes: We should probably pin this to stay between 1.0.0 and 2.0.0

eth_account/_utils/signing.py
```diff
@@ -44,6 +58,175 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+    for field_type in structured_data["types"]:
+        for field in structured_data["types"][field_type]:
+            # Check that field["name"] matches with IDENTIFIER_REGEX
+            if not re.match(IDENTIFIER_REGEX, field["name"]):
+                raise AttributeError(
+                    "Invalid Identifier `{}` in `{}`".format(field["name"], field_type)
+                )
+            # Check that field["type"] matches with TYPE_REGEX
+            if not re.match(TYPE_REGEX, field["type"]):
+                raise AttributeError(
```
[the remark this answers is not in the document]
Bhargavasomu: @kclowes this is actually an `AttributeError` because here the `type` **attribute** won't be valid.

eth_account/_utils/signing.py
```diff
@@ -44,6 +58,175 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+    return field["type"] + " " + field["name"]
+
+
+def encodeType(primaryType, types):
+    # Getting the dependencies and sorting them alphabetically as per EIP712
+    deps = dependencies(primaryType, types)
+    deps_without_primary_type = list(filter(lambda x: x != primaryType, deps))
+    sorted_deps = [primaryType] + sorted(deps_without_primary_type)
+
+    result = ''.join(
```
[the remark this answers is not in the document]
Bhargavasomu: @kclowes I actually don't want to use another function here because, it would be small code so as to encapsulate in a function. Further the reason that I have used `list comprehension` was to speed up the process which would be gone if I were to use a function.
But nevertheless, please let me know if you had any other intention for putting it inside the function. Will change it.

tests/core/test_EIP712.py
Bhargavasomu: I'm in search of such a method.

Bhargavasomu:
Also @kclowes should we move out the `EIP712` functionalities into a separate `_utils` file? Because it seems worth another file.

eth_account/_utils/signing.py
```diff
@@ -44,7 +58,182 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
+            hashed_value = keccak(primitive=encodeData(field["type"], types, value))
+            encValues.append(hashed_value)
+        elif field["type"][-1] == "]":
+            # TODO: Replace the above conditionality with Regex for identifying arrays declaration
+            raise NotImplementedError("TODO: Arrays currently unimplemented in encodeData")
+        else:
+            # First checking to see if the individual values can be encoded
+            try:
+                is_encodable(field["type"], value)
+            except:
```
pipermerriam: Does not appear that `is_encodable` raises exceptions.

https://github.com/ethereum/eth-abi/blob/6aa2bd0e3a5f213afd4d475dc6f0689b31875637/eth_abi/codec.py#L89-L114

Bhargavasomu: @pipermerriam The exception is raised here 
https://github.com/ethereum/eth-abi/blob/6aa2bd0e3a5f213afd4d475dc6f0689b31875637/eth_abi/codec.py#L102.

This is triggered when we give something like `is_encodable("hello", "something")`

pipermerriam: Ah!

I opened https://github.com/ethereum/eth-abi/issues/125to track the upstream issue

For now how about we add an `is_valid_type` utility that **always** returns `True/False`, underlying implementation can depend on the exception raising behavior from `is_encodable`.  Make sure to link to the `eth-abi` issue in that utility function's docstring so that it's clear why it exists and also so we can know when to remove it once that issue has been addressed in `eth-abi`.

Bhargavasomu:
Also I feel that this whole `EIP 712` can have its own `_utils module` which would include the `validation.py` and `hashing.py` as separate files. Am I good to go on that (or) do you feel that putting all of them in the same file is good?

pipermerriam:
I'm fine having an `eth_account._utils.eip712` module.

kclowes:
If it's all the same to you two, I'd rather call it something more descriptive like `structured_data_signing` or something rather than `eip712`

eth_account/_utils/signing.py
```diff
@@ -44,7 +58,182 @@ def sign_transaction_dict(eth_key, transaction_dict):
...
-# watch here for updates to signature format: https://github.com/ethereum/EIPs/issues/191
+#
+# EIP712 Functionalities
+#
+def dependencies(primaryType, types, found=None):
+    """
+    Recursively get all the dependencies of the primaryType
+    """
+    # This is done to avoid the by-reference call of python
+    found = found or []
```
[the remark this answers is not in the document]
Bhargavasomu: @pipermerriam I have replaced the `recursive version` with a `DFS` style approach. Hence now the `dependencies` function is using the `set datastructure` for `inclusion checks`, but in the end while returning, it is being made into a tuple. You can check this out in the latest push. Thankyou.

tests/core/test_EIP712.py
Bhargavasomu: @pipermerriam I was wondering how to go about this. How about we maintain a lot of `JSON` files in the `tests` directory and load them as `fixtures`, instead of writing those JSON strings in the test file? (This is similar to what we are doing in `py-evm` and `trinity`)

pipermerriam: I'm :+1: on using individual JSON files in an appropriately placed `./fixtures/` directory (probably alongside these tests).

eth_account/_utils/signing.py
```diff
@@ -44,7 +49,9 @@ def sign_transaction_dict(eth_key, transaction_dict):
     return (v, r, s, encoded_transaction)
 
 
-# watch here for updates to signature format: https://github.com/ethereum/EIPs/issues/191
+# watch here for updates to signature format:
+# https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md
+# https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md
```
pipermerriam: @kclowes you might go subscribe yourself to the relevant discussions for these eips to be sure we catch any updates that happen.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,212 @@
...
+                )
+            # Special case where the values need to be keccak hashed before they are encoded
+            hashed_value = keccak(primitive=value)
+            yield "bytes32", hashed_value
+        elif field["type"] in types:
+            # This means that this type is a user defined type
+            hashed_value = keccak(primitive=encodeData(field["type"], types, value))
+            yield "bytes32", hashed_value
+        elif field["type"][-1] == "]":
+            # TODO: Replace the above conditionality with Regex for identifying arrays declaration
```
pipermerriam: cc @davesque 

So we have a very nice mechanism in eth-abi to do this but it isn't a public API.  Maybe it should be.

https://github.com/ethereum/eth-abi/blob/master/eth_abi/grammar.py

This approach is quite robust (more so than a regex based approach) and might be worth exploring how we could expose that functionality for use here.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,212 @@
...
+            else:
+                struct_names_yet_to_be_expanded.append(field["type"])
+
+    # Don't need to make a struct as dependency of itself
+    deps.remove(primaryType)
+
+    return tuple(deps)
+
+
+def dict_to_type_name_converter(field):
```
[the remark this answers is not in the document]
Bhargavasomu: @pipermerriam this function doesn't just give the `field name`, but it also includes the `field type`.

So how about we go with `construct_field_representation`? Since what it is returning is the string representation of the `field`.

carver:
> @carver might want to give it a look too since this is the `eth-account` codebase and this stuff is sensitive code in case there are any red flags you see that I dont.

I have some ideas about how I would restructure the code (including some code not added in this PR), and add a new API, but it is not a quick thing so I'll save it for later.

I don't see any red flags in any of the *existing* code. I didn't do a deep dive on the structured message encoder implementation, but it's all pretty separated from any key handling. No objections :+1:

eth_account/_utils/structured_data/hashing.py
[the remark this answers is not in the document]
pipermerriam: does the term `identifier` work well here?

eth_account/_utils/structured_data/hashing.py
davesque: The `eth_abi.grammar` module is public API after some recent changes.

davesque: https://github.com/ethereum/eth-abi/blob/master/docs/grammar.rst

eth_account/_utils/structured_data/hashing.py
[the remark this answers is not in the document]
Bhargavasomu: @pipermerriam how about `field_identifier`? (Because I have used `IDENTIFIER_REGEX` which  checks that `field["name"]` follows a pattern.)

eth_account/_utils/signing.py
carver: cc @davesque in case you hadn't seen this interesting application of a custom eth-abi registry: the typed signed message EIP 712.

davesque: Skimmed over it just now.  Nifty.  Looking forward to seeing how the registry gets used for that.

davesque:
@Bhargavasomu I just cut a new `eth-abi` beta release that fixes the type parsing issue: https://github.com/ethereum/eth-abi/issues/126

Bhargavasomu:
@pipermerriam @kclowes The whole `EIP` has been implemented (Apart from the `rendering changes` that need to be made in `web3.py` side). Could you please review. Thankyou.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,279 @@
...
+    """
+    if not isinstance(data, list) and not isinstance(data, tuple):
+        # Not checking for Iterable instance, because even Dictionaries and strings
+        # are considered as iterables, but that's not what we want the condition to be.
+        return ()
+
+    expected_dimensions = get_array_dimensions(data[0])
+    for index in range(1, len(data)):
+        # 1 dimension less sub-arrays should all have the same dimensions to be a valid array
+        if get_array_dimensions(data[index]) != expected_dimensions:
```
pipermerriam: I see two issues with this.  I think this is a pretty costly operation since it recursively re-computes these array dimensions making this function have something like an `O(N*2)` complexity or worse.  

Maybe you could re-structure this as something like the following and do this in two stages.

```python
@to_tuple
def enumerate_depth_and_dimensions(data, depth):
    """
    Yields 2-tuples of depth and length of element at that depth
    """
    if not is_non_string_iterable(v):
        return
    
    yield depth, len(data)

    for item in data:
        yield from enumerate_depth_and_dimensions(item, depth + 1)

def get_array_dimensions(data):
    depths_and_dimensions = get_depths_and_dimensions(data)
    # re-form as a dictionary with `depth` as key, and all of the dimensions found at that depth.
    grouped_by_depth = toolz.groupby(operator.itemgetter(0), depths_and_dimensions)
    # validate that there is only one dimension for any given depth.
    invalid_depths = tuple(depth for depth, dimensions in grouped_by_depths.items() if len(set(dimensions)) != 1)
    if invalid_depths:
        raise ValidationError("...")
    dimensions = tuple(toolz.first(set(dimensions)) for depth, dimensions in sorted(grouped_by_depth.items()))
```

This is of course untested psuedo-code and should be cleaned up for readabiltity but it should demonstrate the following algorithem.

- walk the full trie producing an enumeration of each element by it's depth and dimension.
- group all of the values together by depth and verify that each depth only has a single dimension, raising an error if this is not true.
- collate the dimension data sorted by depth

The error messaging could be improved if the enumeration of the depths and dimensions also contained the path to that part of the data which would allow for better error reporting.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,279 @@
...
+            raise TypeError("Not a valid array or incomplete array")
+
+    return (len(data),) + expected_dimensions
+
+
+@to_tuple
+def flatten_multidimensional_array(array):
+    for item in array:
+        if isinstance(item, Iterable) and not isinstance(item, str):
+            for x in flatten_multidimensional_array(item):
```
pipermerriam: You can replace this with `yield from flatten_multidimensional_array(item)`

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,279 @@
...
+                        value,
+                        type(value),
+                        field["type"],
+                    )
+                )
+
+
+def encode_data(primaryType, types, data):
+    data_types_and_hashes = _encode_data(primaryType, types, data)
+    data_types, data_hashes = zip(*data_types_and_hashes)
```
pipermerriam: This will fail if `data_types_and_hashes` is empty.  Something to be aware of with this pattern that typically has to be special cased.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,279 @@
...
+    The structured_json_string_data is expected to have the ``types`` attribute and
+    the ``primaryType``, ``message``, ``domain`` attribute.
+    The ``is_domain_separator`` variable is used to calculate the ``hashStruct`` as
+    part of the ``domainSeparator`` calculation.
+    """
+    structured_data = json.loads(structured_json_string_data)
+    validate_structured_data(structured_data)
+
+    types = structured_data["types"]
+    if is_domain_separator:
```
pipermerriam: after reading through this one more time I think this function should be decomposed as follows.  My reasoning is that boolean flags like this are an indication that there is some inner functionality that isn't properly separated, thus requiring a feature flag to be able to re-use that functionality.

Here's where I think the lines should be.

1. Outermost API that the user interacts with is responsible for loading the JSON and validating it.  All remaining functions assume inputs are valid.
2. `hash_domain(structured_data)`:
    - simple wrapper for `keccak(encode_data("EIP712Domain", structured_data["types"], structured_data["domain"]))`
3. `hash_message(structured_data)`:
    - simple wrapper for `keccak(encode_data(structured_data["primaryType"], structured_data["types"], structured_data["message"]))`


Probably good to have some high level nice error messages if any of the `"types"` / `"domain"` / `"message"` / `"primaryType"` keys are missing.

eth_account/_utils/structured_data/hashing.py
Bhargavasomu: @pipermerriam such a case won't occur. This is because `_encode_data` function returns atleast one `2 length tuple`, which would be `("bytes32", hash_struct_type(primary_type, types))`

eth_account/_utils/structured_data/hashing.py
Bhargavasomu: @pipermerriam I don't think there is a **difference** in the `time complexity` of my approach when compared to yours. But the approach that you suggested is easy to understand. Hence I will change it.

Bhargavasomu: Ah my bad, I see where your approach works better.

Bhargavasomu:
@pipermerriam @kclowes review please.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,319 @@
...
+    abi_type = parse(type)
+    return abi_type.is_array
+
+
+@to_tuple
+def get_depths_and_dimensions(data, depth):
+    """
+    Yields 2-length tuples of depth and dimension of each element at that depth
+    """
+    if not isinstance(data, list) and not isinstance(data, tuple):
```
pipermerriam: This should be collapsed to a single check using `isinstance(data, (list, tuple))`.  If you want you can also use `eth_utils.is_list_like` though it is a bit more forgiving, allowing *any* `collections.abc.Sequence` type that is not a string type which probably isn't what we want here.

eth_account/_utils/structured_data/hashing.py
```diff
@@ -0,0 +1,319 @@
...
+def get_array_dimensions(data):
+    """
+    Given an array type data item, check that it is an array and
+    return the dimensions as a tuple.
+    Ex: get_array_dimensions([[1, 2, 3], [4, 5, 6]]) returns (2, 3)
+    """
+    depths_and_dimensions = get_depths_and_dimensions(data, 0)
+    # re-form as a dictionary with `depth` as key, and all of the dimensions found at that depth.
+    grouped_by_depth = {
+        depth: tuple(tuple(zip(*group))[1])
```
pipermerriam: Do you think either of the following are more legible?

- `tuple(map(operator.itemgetter(1), group))`
- `tuple(dimension for depth, dimension in group)`  (I think this is my preference and I *think* it might be the most performant)
```

File: eth_account/_utils/structured_data/hashing.py [hunk 18]
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    # validate that there is only one dimension for any given depth.
+    invalid_depths = tuple(
+        depth
+        for depth, dimensions in grouped_by_depth.items()
+        if len(set(dimensions)) != 1
+    )
+    if invalid_depths:
+        raise ValidationError("Not a valid array or incomplete array")
```
pipermerriam: I think we should do the extra work to construct an informative error message here that includes each of the invalid depths and what dimensions were found at that depth.  Otherwise a user confronted with this error message is unlikely to have any idea what went wrong without a deep understanding of the signing spec which I think is not likely.

Bhargavasomu:
@pipermerriam I have addressed all your issues. Thankyou.

davesque:
I think this was requested here.  Just added some new methods to the `eth-abi` API to allow for checking if a type is recognized by the registry.  Cut `eth-abi` v2.0.0-beta.8 just now which includes those features.

davesque:
Method documentation found here:
https://eth-abi.readthedocs.io/en/latest/eth_abi.html#eth_abi.codec.ABIEncoder.is_encodable_type
https://eth-abi.readthedocs.io/en/latest/eth_abi.html#eth_abi.registry.ABIRegistry.has_encoder

Related exceptions (you probably won't have to worry about these too much):
https://eth-abi.readthedocs.io/en/latest/eth_abi.html#eth_abi.exceptions.NoEntriesFound
https://eth-abi.readthedocs.io/en/latest/eth_abi.html#eth_abi.exceptions.MultipleEntriesFound

Bhargavasomu:
@davesque I don't think the new version of `eth-abi` which is `2.0.0b8` is found by `python 3.5`. It's recognized only by `python 3.6`

davesque:
@Bhargavasomu Correct.  Python 3.5 support was removed.

Bhargavasomu:
@davesque but here in `eth-account`, we are supporting both `python 3.5 and pypy`. 

@pipermerriam @carver should we remove the support for `python 3.5 and pypy`?

pipermerriam:
@Bhargavasomu we're discussing internally what we want to do about this python issue.  Maybe you can revert your last changes and keep this PR using it's own hand rolled utility and open an issue to track the need to update once python3.5 is no longer supported.

Bhargavasomu:
This PR is now back to the old version and have opened the issue here https://github.com/ethereum/eth-account/issues/59

Bhargavasomu:
Is there something more to be done as part of this PR? If not, could someone please merge this?

PR #60, opened 2019-04-25
Project template upgrade

carver:
Upgrades from the project template

#### Cute Animal Picture


carver:
Hm, maybe it's time to drop rlp0 and py3.5 (and pypy3) support. `eth-account` is v0, so it would still only be a minor (but breaking) bump.

pipermerriam:
dropping 3.5 seems viable.   Web3.py is correctly pinned to `<0.4.0` which is the main place this would be an issue.

pipermerriam:
dropping pypy across the board also seems prudent.

PR #61, opened 2019-04-26
New Message Signing API

### LINKED ISSUE #58: New message signing API
## What was wrong?

Message signing should be very pluggable according to EIP191. Message encoding is all built-in to the (now confusingly named) `defunct_hash_message()`.

## How can it be fixed?

New message signing API that captures this concept:
```py

class SignableMessage(NamedTuple):  # as defined by EIP-191
  version: HexBytes  # must be length 1
  header: HexBytes  # aka "version specific data"
  body: HexBytes  # aka "data to sign"

signable_message = encode_structured_data_message(json_str)
account.sign_message(signable_message)
```

The encoding method could have arbitrary inputs without changing the message signing API, like:
```py
signable_message = encode_intended_validator(message, validator_address)
account.sign_message(signable_message)
```

It abstracts away the EIP-191 encoding process and message hashing inside `sign_message()`. A new `recover_message()` could work similarly.

The solution should make it trivial to experiment with or produce your own (sub)version of EIP-191, without any change to the eth-account library. (This is already possible by signing and recovering hashes of your own making, but it's not as accessible).

carver:
## What was wrong?

Fixes #58 

## How was it fixed?

Roughly as defined in #58 -- add a new `SignableMessage` built on the EIP-191 concept.

Add a few `encode_*` methods to encode different types of messages into `SignableMessage`s

- [x] More documentation to do
- [x] Linting
- [x] mark/document `hash_eip191_message` as private
- [x] squash

Note that this also:
- removes an old link in `eth_account.local` that was just for backwards compatibility (next release is going to have a breaking-change bump)
- introduces some f-string formatting since we no longer support py3.5
- introduces some snake-casing (I plan to open a separate PR to add snake casing to all the APIs, and add deprecation warnings to the others. The camel-case versions can be removed in v0.5

#### Cute Animal Picture


tests/core/test_validation.py
```diff
@@ -15,7 +15,7 @@
     'nonce': 3,
 }
 
-TEST_PRIVATE_KEY = b'\0' * 32
+TEST_PRIVATE_KEY = b'\0' * 31 + b'\x01'
```
carver: Newer crypto libraries complain about using an all-zero private key (yay!). We still use a very unsafe key for tests, but not 0.

carver:
Ping @pipermerriam for review

carver:
Also, the last two commits of #64 which is built on top of this one.

eth_account/account.py
```diff
@@ -361,50 +360,38 @@ def setKeyBackend(self, backend):
...
 
         .. code-block:: python
 
             >>> msg = "I♥SF"
-            >>> from eth_account.messages import defunct_hash_message
-            >>> msghash = defunct_hash_message(text=msg)
-            HexBytes('0x1476abb745d423bf09273f1afd887d951181d25adc66c4834a70491911b7f750')
+            >>> from eth_account.messages import encode_defunct
+            >>> msghash = encode_defunct(text=msg)
+            SignableMessage(version=b'E', header=b'thereum Signed Message:\n6', body=b'I\xe2\x99\xa5SF')
```
davesque: Should we maybe add a note about where that version "number" comes from?  Seems like it could cause confusion or people might incorrectly think there's a typo in the documentation.

eth_account/messages.py
```diff
@@ -1,90 +1,232 @@
...
+
+# watch for updates to signature format
+class SignableMessage(NamedTuple):
+    """
+    These are the components of an EIP-191_ signable message. Other message formats
+    can be encoded into this format for easy signing. This data structure doesn't need to
+    know about the original message format.
+
+    In typical usage, you should never need to create these by hand. Instead, use
+    one of the available encode_* methods in this module, like:
```
davesque: Maybe use monospace formatting here?
```suggestion
    one of the available ``encode_*`` methods in this module, like:
```

tests/core/test_accounts.py
```diff
@@ -94,12 +100,17 @@ def PRIVATE_KEY_ALT(request):
 
 
 @pytest.fixture
-def acct(request):
+def acct():
```
davesque: Is there much point to having this fixture?  Seems to just return a reference to a class that's already in scope.

setup.py
```diff
@@ -7,6 +7,7 @@
 
 extras_require = {
     'test': [
+        "hypothesis>=4.18.0,<5",
         "pytest>=3.6.0",
         "pytest-xdist",
```
njgheorghita: Just a note. The newest `pytest-xdist` version `1.28.0` requires `pytest>=4.4.0`. It's fine here since pytest isn't constrained like `"pytest>=3.6.0,<4"`, and `pytest` version `4.4.0` gets installed, but it might be worth updating explicitly

eth_account/account.py
```diff
@@ -222,13 +227,12 @@ def privateKeyToAccount(self, private_key):
         return LocalAccount(key, self)
 
     @combomethod
-    def recoverHash(self, message_hash, vrs=None, signature=None):
+    def recover_message(self, signable_message: SignableMessage, vrs=None, signature=None):
```
njgheorghita: What's the reason for using `recover_message` as the function name? I don't have the most complete knowledge of all the cryptography terminology, but something like `recover_address` is more intuitive to someone like me

tests/core/test_accounts.py
```diff
@@ -280,6 +293,47 @@ def test_eth_account_hash_message_hexstr(acct, message, expected):
...
+
+    # sign via hash
+    msg_hash_hex = defunct_hash_message(hexstr=message_hex)
+    signed_via_hash_hex = keyed_acct.signHash(msg_hash_hex)
+
+    # sign via message
+    signable_message_hex = encode_defunct(hexstr=message_hex)
+    signed_via_message_hex = keyed_acct.sign_message(signable_message_hex)
+
+    assert signed_via_hash_hex == signed_via_message_hex
```
njgheorghita: Oooh, I like these tests 😄

eth_account/messages.py
carver: :confused: it doesn't feel quite right. I think it's because it feels wrong to use monospace for anything that's not a proper explicit reference. Since this sort of obliquely references the methods (with the `*`), it doesn't feel right. I could maybe add a bullet list with a couple examples that are referenced explicitly.

tests/core/test_accounts.py
carver: Hm, yeah, I think it used to supply both the instance and the class... I guess that got dropped at some point. Investigating...

carver: Looks like it still works as a class or instance, when I added that back in, so I did.

eth_account/messages.py
```diff
@@ -1,90 +1,232 @@
...
+
+    :param primitive: the binary message to be signed
+    :type primitive: bytes or int or Mapping (eg~ dict )
+    :param hexstr: the message encoded as hex
+    :param text: the message as a series of unicode characters (a normal Py3 str)
+    :returns: The EIP-191 encoded message, ready for signing
+
+    .. _EIP-712: https://eips.ethereum.org/EIPS/eip-712
+    """
+    if isinstance(primitive, Mapping):
```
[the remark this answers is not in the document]
carver: There's no reason it needs to be mutable, that I can think of. It's just showing intention for broader support than only the specific `Dict` class. We might want to support `AttributeDict`, for example.

eth_account/messages.py
```diff
@@ -1,90 +1,232 @@
...
+        message_bytes,
     )
-    return recovery_hasher(message_bytes)
+
+
+def defunct_hash_message(
+        primitive: bytes = None,
+        *,
+        hexstr: str = None,
+        text: str = None) -> SignableMessage:
```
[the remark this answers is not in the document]
carver: Just a typo, thanks!

eth_account/account.py
carver: Right, so this is "running `recover` on a message", not "recover the message". I can definitely see how if you're not swimming in `ecrecover` all day it might not be obvious. Hopefully the docs explain what it does enough to avoid confusion. Whether you are running recover on the transaction or on the message, you are always recovering an address, so `recover_address` is ambiguous. `recover_on_message` feels pretty unwieldy to me. I don't really have any better ideas.

eth_account/messages.py
davesque: I guess I had always used single backtick links for explicit references. The double backtick mono-space formatting seems like it's meant for a broader range of situations and might include this case. But whatevs. No biggie.

PR #63, opened 2019-04-27
Make Account.default_kdf private (and release notes)

carver:
## What was wrong?

It doesn't seem like we need so many different ways to set a default kdf. An environment variable, and a keyword arg on `encrypt()` seem sufficient.  If we find more use cases, then adding it back as public is a lot easier than removing it later.

Also, added a bunch of release notes ahead of the v0.4.0 release.

## How was it fixed?

- Added `_` in front.
- Moved docs about kdf environment variable into the method where you would find it relevant (`encrypt`).
- Added release notes

I expect a cryptography failure which will be resolved after rebasing on #61 

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -144,7 +144,7 @@ class Account(object):
         :returns: The data to use in your encrypted file
         :rtype: dict
 
-        If kdf is not set, the default key deriviation function falls back to the
+        If kdf is not set, the default key derivation function falls back to the
         environment variable :envvar:`ETH_ACCOUNT_KDF`. If that is not set, then
         'scrypt' will be used as the default.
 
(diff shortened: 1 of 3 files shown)


carver:
(worth mentioning that `Account.default_kdf` was never released, so it's not a notable change in the release notes)

eth_account/account.py
```diff
@@ -149,6 +144,10 @@ def encrypt(cls, private_key, password, kdf=None, iterations=None):
         :returns: The data to use in your encrypted file
         :rtype: dict
 
+        If kdf is not set, the default key deriviation function falls back to the
```
davesque:
```suggestion
        If kdf is not set, the default key derivation function falls back to the
```

PR #64, opened 2019-04-29
Deprecate camel-case

carver:
## What was wrong?

There are camelCase APIs on eth-account, plus some names that are longer than they need to be.

## How was it fixed?

Deprecated old methods, added snake case methods. Shortened a few:

- `Account.privateKeyToAccount()' -> 'Account.from_key()'
- `LocalAccount.privateKey` -> `LocalAccount.key`

Also, fixed the pydocstyle run, and resolved all the related warnings.

This is blocked on merging #61 . Only the last two commits are not part of #61: 
- 54b8033
- 1b8979c

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -80,16 +80,16 @@ class Account(object):
             >>> acct = Account.create('KEYSMASH FJAFJKLDSKF7JKFDJ 1530')
             >>> acct.address
             '0x5ce9454909639D2D17A3F753ce7d93fa0b9aB12E'
-            >>> acct.privateKey
+            >>> acct.key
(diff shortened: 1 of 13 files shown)


eth_account/account.py
```diff
@@ -56,23 +56,23 @@
 
 
 class Account(object):
-    '''
-    This is the primary entry point for working with Ethereum private keys.
+    """
+    Account is the primary entry point for working with Ethereum private keys.
```
davesque: Pretty nitpicky but...maybe just remove "Account is" from the beginning and capitalize the "t" in "the"?  Or, use the double-backtick monospace formatting for signify that "Account" refers to a class and isn't just another word that begins the sentence.  But I don't like that because then the name of the class could potentially get out of sync with the documentation.

Issue #59, opened 2019-04-01
Replace is_valid_abi_type with ABI's is_encodable_type method


Bhargavasomu:
## What was wrong?
Currently `eth-account` is still supporting `python 3.5`. But to replace utils method `is_valid_abi_type` with ABI's method `is_encodable_type`, it is possible only when the support of `python 3.5` is removed by `eth-account`, because `eth-abi` currently doesn't support `python 3.5`

## How can it be fixed?

Replace the function, when the support for `python 3.5` is dropped in `eth-account`.

Bhargavasomu:
Further details related to this issue can be found [here](https://github.com/ethereum/eth-account/pull/57#issuecomment-478207050)

carver:
Closed by #67

PR #67, opened 2019-05-31
Use is_encodable_type

fubuloubu:
## What was wrong?
Replace `is_valid_abi_type` with `is_encodable_type` from `eth-abi` when Python 3.5 support was dropped from this repository

Fixes Issue #59

#### Cute Animal Picture
baby tiger

### DIFF
--- a/tests/core/test_structured_data_signing.py
+++ b/tests/core/test_structured_data_signing.py
@@ -2,6 +2,9 @@ import json
 import pytest
 import re
 
+from eth_abi.exceptions import (
+    ABITypeError,
+)
 from eth_utils import (
     ValidationError,
     keccak,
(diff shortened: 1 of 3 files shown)


tests/core/test_structured_data_signing.py
```diff
@@ -233,9 +236,9 @@ def test_invalid_structured_data_invalid_abi_type():
         "tests/fixtures/invalid_message_invalid_abi_type.json"
     ).read()
     invalid_structured_data = json.loads(invalid_structured_data_string)
-    with pytest.raises(TypeError) as e:
+    with pytest.raises(ABITypeError) as e:
```
carver: Could use the `match` keyword instead of the `assert`: https://docs.pytest.org/en/latest/assert.html

(Obviously just an optional suggestion to fix broken windows, since this pattern was already here pre-PR)

fubuloubu: Giving it a shot

fubuloubu: Nope, had some weird issue.

fubuloubu:
Sqaushed it

PR #76, opened 2019-12-20
Replace attrdict with NamedTuple

cburgdorf:
## What was wrong?

The usage of `attrdict` dependency was causing a deprecation warning.

```
  from collections import (
/home/cburgdorf/Documents/hacking/ef/trinity/venv/lib/python3.7/site-packages/attrdict/mixins.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

```

## How was it fixed?

Replaced with `NamedTuple` which should have the positive side effect of a stronger guarantee of immutability.

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -44,7 +44,8 @@ from eth_account._utils.transactions import (
     vrs_from,
 )
 from eth_account.datastructures import (
-    AttributeDict,
+    SignedMessage,
+    SignedTransaction,
 )
 from eth_account.hdaccount import (
     ETHEREUM_DEFAULT_PATH,
@@ -517,7 +518,7 @@ class Account(object):
         :param private_key: the key to sign the message with
         :type private_key: hex str, bytes, int or :class:`eth_keys.datatypes.PrivateKey`
         :returns: Various details about the signature - most importantly the fields: v, r, and s
-        :rtype: ~eth_account.datastructures.AttributeDict
+        :rtype: ~eth_account.datastructures.SignedMessage
 
         .. doctest:: python
 
@@ -531,12 +532,13 @@ class Account(object):
             >>> # If you're curious about the internal fields of SignableMessage, take a look at EIP-191, linked above  # noqa: E501
             >>> key = "0xb25c7db31feed9122727bf0939dc769a96564b2de4c4726d035b36ecf1e5b364"
             >>> Account.sign_message(msghash, key)
-            AttrDict({'messageHash':
-             HexBytes('0x1476abb745d423bf09273f1afd887d951181d25adc66c4834a70491911b7f750'),
-             'r': 104389933075820307925104709181714897380569894203213074526835978196648170704563,
-             's': 28205917190874851400050446352651915501321657673772411533993420917949420456142,
-             'v': 28,
-             'signature': HexBytes('...')})
+            SignedMessage(messageHash=HexBytes('0x1476abb745d423bf09273f1afd887d951181d25adc66c4834a70491911b7f750'),
+             r=104389933075820307925104709181714897380569894203213074526835978196648170704563,
+             s=28205917190874851400050446352651915501321657673772411533993420917949420456142,
+             v=28,
+             signature=HexBytes('0xe6ca9bba58c88611fad66a6ce8f996908195593807c4b38bd528d2cff09d4eb33e5bfbbf4d3e39b1a2fd816a7680c19ebebaf3a141b239934ad43cb33fcec8ce1c'))
+
+
 
         .. _EIP-191: https://eips.ethereum.org/EIPS/eip-191
         """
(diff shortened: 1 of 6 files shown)


cburgdorf:
review ping @pipermerriam

cburgdorf:
>It doesn't appear that this is fixed upstream, nor has there been a release in about 11 months.

The inactivity there was part of the reason why I believe it is not worth to try to get this fixed upstream. I wouldn't expect to this to result in an upstream release in a short time frame which is why I thought it is better to fix it on our on end. Especially if we consider that modern Python has no need for that library any more.

>Second option is to use something like this and vendor it for the time being: https://github.com/ethereum/web3.py/blob/c1c80dba4d9c203af7a4963bf1ce5e922f1172da/web3/datastructures.py

But wouldn't that be as much of a breaking change as just going with `NamedTuple`? I'm assuming the breaking part is just the underlying object that is being returned and maybe the `repr_pretty` which I could still put back in. Other than that, the solution proposed in the PR is attribute access compatible.

Someone calling `signTransaction` can still access `signed_tx.rawTransaction` or any other attribute like before. I might be missing some other subtle breaking change?

>I'm good with us eventually switching to a NamedTuple but I consider this library to be somewhat foundational and thus a major version bump in it is going to need to be taken carefully and probably shouldn't happen until sometime in January when all this holiday nonsense has passed.

Yeah, no rush. I'd probably say, let's just wait and cut a new major release some time in January.

pipermerriam:
It is a breaking change because named tuples don't support dictionary style key retrieval.  Whatever we return needs to support **both** attribute and string-based key lookups via the standard `Mapping` interface to retain backwards compat.

cburgdorf:
>string-based key lookups

Ah! That makes sense. Thanks for clarifying.

veox:
2c: probably means there needs to be a test for this.

cburgdorf:
>There will be no release: the ustream repo has been archived by the owner.

Considering that the project seems dead, I'd argue we should just cut a new major release and migrate Trinity to it. We can continue to backport further changes into the `0.x` development line if the community demands it, but considering that this breaking change is really tiny and easy to resolve I do not actually see that even becoming an issue.

cburgdorf:
I have updated the change to preserve the current behavior of providing index access. I also tuned the tests to cover indexing access. What do you think about this @pipermerriam ?

I'd love to get this fixed as it's currently the main driver of nasty warnings in Trinity.

pipermerriam:
@carver any thoughts on the breaking change here.  I'm inclined to :shipit: since we're still in the `0.x` release line and thus I believe we're allowed to introduce breaking changes in minor point releases.

cburgdorf:
@pipermerriam what's the remaining breaking change now that I added index access support?

eth_account/account.py
```diff
@@ -507,7 +508,7 @@ def sign_message(self, signable_message: SignableMessage, private_key):
         :param private_key: the key to sign the message with
         :type private_key: hex str, bytes, int or :class:`eth_keys.datatypes.PrivateKey`
         :returns: Various details about the signature - most importantly the fields: v, r, and s
-        :rtype: ~eth_account.datastructures.AttributeDict
+        :rtype: ~eth_account.datastructures.SignedMessage
```
carver, line 511: Interesting that the doctests (which were supposedly checked recently) didn't catch that the returned value changes below.

I would expect there to be a number of places in the docs where this would change.

carver:
> @pipermerriam what's the remaining breaking change now that I added index access support?

Well, for very type-conscious consumers, it could still break their tests. (Including things like Web3.py?) Maybe their lint tests or doctests.

I have always preferred a more duck-typing approach, but with the strict-typing tactic used in a lot of our projects, we might still be forced to do it is a v0 minor release (which implies a breaking change).

I still think it's a good change to keep the indexing access to minimize the cost of updating downstream projects. :+1:

eth_account/account.py
cburgdorf: I just spent 1,5 hour to depuzzle this and here's the best I know so far:

1. We never had any doctest running in this repo so far. I wasn't able to find a single CI run that ran any doctests (even though they exist)

2. Then I found https://github.com/ethereum/eth-account/pull/94 which seems to fix this (it reports running 50 tests)

I'd happily rebase this on top of #94 but in case #94 needs more time to finish it should be easy to rebase that one on top of these changes, too.

kclowes: 🤦 I totally lost track of #94. I just updated and it should be ready to merge. @cburgdorf I'm happy to pull your changes into #94, or vice versa, whatever is easier! Just let me know.

cburgdorf: @kclowes I don't have merge rights in this repo. If you have merge rights then feel free to merge both PRs in any preferred order.

pipermerriam:
@cburgdorf I merged the other PR, mind rebasing to make sure everything is still good here and then I'll get this merged.

cburgdorf:
@pipermerriam rebased and ready to :ship:

pipermerriam:
@carver can you handle the release?

carver:
Seems like a fuzzy gray area of incompatibility. I'll just be sure to highlight it in the release notes. I just checked web3.py, and it doesn't have any reference to `AttributeDict` being returned by eth-account AFAICT.

So I'll go ahead and release this as a point release.

carver:
It's a little late today, so I'll aim for Thursday morning.

PR #78, opened 2019-12-27
Setup coverage testing via pytest

AndreMiras:
Follow-up PR can publish it to coveralls.io for better tracking.
Note that current coverage is 88%.

## What was wrong?

We currently have no track of the library coverage.

## How was it fixed?

Simply adding `pytest-cov` test dependency and adding the `--cov` flag when `pytest` is ran.

#### Cute Animal Picture

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ lint-roll:
 	$(MAKE) lint
 
 test:
-	pytest tests
+	pytest tests --cov eth_account/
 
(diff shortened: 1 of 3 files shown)


pacrob:
Too old! Reviving discussion in PR #192

AndreMiras:
I could have rebased and updated, but great that created one then

PR #80, opened 2020-01-05
Fix typo in SignableMessage docstring

jstoxrocky:
## What was wrong?
I was trying to wrap my head around the EIPs about signing messages and found a small typo in the SignableMessage docstring. Probably meant to refer to the third method of encoding: `encode_defunct`.

#### Cute Animal Picture


kclowes:
Thanks for the PR! Sorry we missed this. Looks like it was fixed in #127.

PR #84, opened 2020-01-18
Fixes encrypt(), decrypt() and messages encode_defunct() doctests

### LINKED ISSUE #81: Doctest Account encrypt()/decrypt()
This is a follow up for https://github.com/ethereum/eth-account/pull/49
`Account.encrypt()` and `Account.decrypt()` doctests got disabled because of `getpass()`. The idea is to write a doctest hardcoding the password to demonstrate both features in the doctest, refs @carver comments in https://github.com/ethereum/eth-account/pull/49#discussion_r363462283 and https://github.com/ethereum/eth-account/pull/49#pullrequestreview-338868887
Apart from these 2 doctests 7 others got skipped (`doctest: +SKIP`) because they broke for different reason. We could also try investigating/fixing them in this task.

AndreMiras:
This is a follow up for comments to PR #49, closes #81.

## What was wrong?
Some doctests from `account.Account.encrypt()`, `account.Account.decrypt()` and `messages.encode_defunct()` were disabled (`+SKIP`) because they were failing.

Issue #81 

## How was it fixed?

- `account.Account.encrypt()` is now providing full deterministic input, including the input password
- `account.Account.decrypt()`
  - leverages [doctest ELLIPSIS](https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS) for non deterministic output
  - uses `pprint()` for clean indentation matching and deterministic dictionary key ordering
- `messages.encode_defunct()` uses `to_hex()` & `to_bytes()` from `eth_utils` rather than `web3`

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -108,19 +108,19 @@ class Account(object):
             >>> encrypted = {
             ... 'address': '5ce9454909639d2d17a3f753ce7d93fa0b9ab12e',
             ... 'crypto': {'cipher': 'aes-128-ctr',
-            ...  'cipherparams': {'iv': '78f214584844e0b241b433d7c3bb8d5f'},
-            ...  'ciphertext': 'd6dbb56e4f54ba6db2e8dc14df17cb7352fdce03681dd3f90ce4b6c1d5af2c4f',
-            ...  'kdf': 'pbkdf2',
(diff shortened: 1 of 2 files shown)


pipermerriam:
Thank you very much for this cleanup.

PR #87, opened 2020-02-01
Add API to Account for mnemonic seed phrases

fubuloubu:
Fixes: #8 #24 
Enables: #25 #37 

## How was it fixed?
Looked at the entire history of attempts at implementing this feature (#33 #34 etc...). Referencing those attempts, I lifted https://github.com/trezor/python-mnemonic (so as not to add a dependency) and implemented BIP32 (`eth_account/hdaccount/deterministic.py`) in a more readable fashion so it is specifically helpful to our scenario, and can be reused (currently only support private parent -> private child, but can be adopted to support public parent -> public child for #25 & #37)

#### Cute Animal Picture
studious monkey

Issue #37, opened 2018-10-10
Add TrezorAccount type


pipermerriam:
STUB for a new account type to support trezor hardware wallets.  Here's a starting point from the shadowlands app:  https://github.com/kayagoban/shadowlands-core/blob/a82c2b32ed46464361d9be10ae9bd6a93039b462/shadowlands/credstick/trezor_ethdriver.py

PR #88, opened 2020-02-03
Add support for BIP32 (Heirarchical Deterministic Wallets)

fubuloubu:
Split off commits related to BIP32 implemented from #87

### DIFF
--- a/eth_account/hdaccount/_utils.py
+++ b/eth_account/hdaccount/_utils.py
@@ -1,0 +1,27 @@
+import hashlib
+import hmac
+
+from eth_keys import (
+    keys,
+)
+from hexbytes import (
+    HexBytes,
+)
+
+SECP256K1_N = int("FFFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFE_BAAEDCE6_AF48A03B_BFD25E8C_D0364141", 16)
+
+
+def hmac_sha512(chain_code: bytes, data: bytes) -> bytes:
+    """
+    As specified by RFC4231 - https://tools.ietf.org/html/rfc4231
+    """
+    return hmac.new(chain_code, data, hashlib.sha512).digest()
+
+
+def ec_point(pkey: bytes) -> bytes:
+    """
+    Compute `point(p)`, where `point` is ecdsa point multiplication
+
+    Note: Result is ecdsa public key serialized to compressed form
+    """
+    return keys.PrivateKey(HexBytes(pkey)).public_key.to_compressed_bytes()
--- a/eth_account/hdaccount/deterministic.py
+++ b/eth_account/hdaccount/deterministic.py
@@ -116,7 +116,7 @@ def derive_child_key(
 
     - Check whether i ≥ 2**31 (whether the child is a hardened key).
       - If so (hardened child):
-        let I = HMAC-SHA512(Key = c_par, Data = 0x00 || ser_256(kpar) || ser_32(i)).
+        let I = HMAC-SHA512(Key = c_par, Data = 0x00 || ser_256(k_par) || ser_32(i)).
         (Note: The 0x00 pads the private key to make it 33 bytes long.)
       - If not (normal child):
         let I = HMAC-SHA512(Key = c_par, Data = ser_P(point(k_par)) || ser_32(i)).
@@ -130,11 +130,11 @@ def derive_child_key(
     assert len(parent_chain_code) == 32
     if isinstance(node, HardNode):
         # NOTE Empty byte is added to align to SoftNode case
-        assert len(parent_key) == 32  # Should be guarenteed here in return statment
+        assert len(parent_key) == 32  # Should be guaranteed here in return statment
         child = hmac_sha512(parent_chain_code, b"\x00" + parent_key + node.serialize())
 
     else:
-        assert len(ec_point(parent_key)) == 33  # Should be guarenteed by Account class
+        assert len(ec_point(parent_key)) == 33  # Should be guaranteed by Account class
         child = hmac_sha512(parent_chain_code, ec_point(parent_key) + node.serialize())
 
     assert len(child) == 64
--- a/eth_account/hdaccount/deterministic.py
+++ b/eth_account/hdaccount/deterministic.py
@@ -8,22 +8,28 @@ BIP-0044: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
 
 Skips serialization and public key derivation as unnecssary for this library's purposes.
 
-    Notes:
-    - Integers are modulo the order of the curve (referred to as n).
-    - Addition (+) of two coordinate pair is defined as application of the EC group operation.
-    - Concatenation (||) is the operation of appending one byte sequence onto another.
+Notes
+-----
 
-    Definitions:
-    - point(p): returns the coordinate pair resulting from EC point multiplication
-      (repeated application of the EC group operation) of the secp256k1 base point
-      with the integer p.
-    - ser_32(i): serialize a 32-bit unsigned integer i as a 4-byte sequence,
-      most significant byte first.
-    - ser_256(p): serializes the integer p as a 32-byte sequence, most significant byte first.
-    - ser_P(P): serializes the coordinate pair P = (x,y) as a byte sequence using SEC1's compressed
-      form: (0x02 or 0x03) || ser_256(x), where the header byte depends on the parity of the
-      omitted y coordinate.
-    - parse_256(p): interprets a 32-byte sequence as a 256-bit number, most significant byte first.
+* Integers are modulo the order of the curve (referred to as n).
+* Addition (+) of two coordinate pair is defined as application of the EC group operation.
+* Concatenation (||) is the operation of appending one byte sequence onto another.
+
+
+Definitions
+-----------
+
+* point(p): returns the coordinate pair resulting from EC point multiplication
+  (repeated application of the EC group operation) of the secp256k1 base point
+  with the integer p.
+* ser_32(i): serialize a 32-bit unsigned integer i as a 4-byte sequence,
+  most significant byte first.
+* ser_256(p): serializes the integer p as a 32-byte sequence, most significant byte first.
+* ser_P(P): serializes the coordinate pair P = (x,y) as a byte sequence using SEC1's compressed
+  form: (0x02 or 0x03) || ser_256(x), where the header byte depends on the parity of the
+  omitted y coordinate.
+* parse_256(p): interprets a 32-byte sequence as a 256-bit number, most significant byte first.
+
 """
 # Additional notes:
 # - This algorithm only implements private parent key → private child key CKD function,
@@ -33,18 +39,19 @@ Skips serialization and public key derivation as unnecssary for this library's p
 #   not intended to be ultimately used for Bitcoin key derivations. This presents a simplified API.
 import hashlib
 import hmac
+from typing import (
+    Tuple,
+    Union,
+)
+
+from eth_keys import (
+    keys,
+)
 from eth_utils import (
     to_int,
 )
 from hexbytes import (
     HexBytes,
-)
-from typing import (
-    Tuple,
-    Union,
-)
-from eth_keys import (
-    keys,
 )
 
 SECP256K1_N = int("FFFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFE_BAAEDCE6_AF48A03B_BFD25E8C_D0364141", 16)
@@ -170,7 +177,7 @@ def derive_child_key(
 
 class HDPath:
     def __init__(self, path: str):
-        '''
+        """
         Constructor for this class. Initializes an hd account generator and
         if possible the encoded key and the derivation path. If no arguments are
         specified, create a new account with createAccount(...) or initialize
@@ -180,7 +187,7 @@ class HDPath:
                                   where idx_* is either an integer value (soft node)
                                   or an integer value followed by either the "'" char
                                   or the "H" char (hardened node)
-        '''
+        """
         nodes = path.split('/')
         if not nodes[0] == 'm':
             raise ValueError(f'Path is not valid: "{path}". Must start with "m"')
--- a/eth_account/hdaccount/deterministic.py
+++ b/eth_account/hdaccount/deterministic.py
@@ -168,7 +168,7 @@ class HDPath:
                                   or the "H" char (hardened node)
         """
         nodes = path.split('/')
-        if not nodes[0] == 'm':
+        if nodes[0] not in ('m', 'M'):
             raise ValueError(f'Path is not valid: "{path}". Must start with "m"')
         decoded_path = []
         try:
--- a/eth_account/hdaccount/deterministic.py
+++ b/eth_account/hdaccount/deterministic.py
@@ -114,7 +114,7 @@ def derive_child_key(
     The function CKDpriv((k_par, c_par), i) → (k_i, c_i) computes a child extended
     private key from the parent extended private key:
 
-    - Check whether i ≥ 231 (whether the child is a hardened key).
+    - Check whether i ≥ 2**31 (whether the child is a hardened key).
       - If so (hardened child):
         let I = HMAC-SHA512(Key = c_par, Data = 0x00 || ser_256(kpar) || ser_32(i)).
         (Note: The 0x00 pads the private key to make it 33 bytes long.)
--- a/eth_account/hdaccount/deterministic.py
+++ b/eth_account/hdaccount/deterministic.py
@@ -68,8 +68,8 @@ class Node(int):
     def __new__(cls, index):
         if 0 > index or index > 2**31:
             raise ValidationError(
-                    f"{cls} cannot be initialized with value {index}"
-                )
+                f"{cls} cannot be initialized with value {index}"
+            )
         obj = int.__new__(cls, index + cls.OFFSET)
         obj.index = index
         return obj
@@ -205,8 +205,8 @@ class HDPath:
                 decoded_path.append(Node.decode(node))
             except ValidationError as err:
                 raise ValidationError(
-                        f'Path "{path}" is not valid. Issue with node "{node}": {err}'
-                    ) from err
+                    f'Path "{path}" is not valid. Issue with node "{node}": {err}'
+                ) from err
 
         self._path = decoded_path
 
--- a/eth_account/hdaccount/deterministic.py
+++ b/eth_account/hdaccount/deterministic.py
@@ -152,9 +152,12 @@ def derive_child_key(
         assert len(parent_key) == 32  # Should be guaranteed here in return statment
         child = hmac_sha512(parent_chain_code, b"\x00" + parent_key + node.serialize())
 
-    else:
+    elif isinstance(node, SoftNode):
         assert len(ec_point(parent_key)) == 33  # Should be guaranteed by Account class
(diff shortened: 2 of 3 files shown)


pipermerriam:
@kclowes will you give this a first pass?

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,217 @@
...
+def ec_point(pkey: bytes) -> bytes:
+    """
+    Compute `point(p)`, where `point` is ecdsa point multiplication
+
+    Note: Result is ecdsa public key serialized to compressed form
+    """
+    return keys.PrivateKey(HexBytes(pkey)).public_key.to_compressed_bytes()
+
+
+class SoftNode(int):
```
kclowes: I don't have a strong opinion about this, but you could DRY up the HardNode and SoftNode a little if you wanted by having them both inherit from a `Node` class or something similar.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,217 @@
...
+def decode_node(node: str) -> Union[SoftNode, HardNode]:
+    if len(node) < 1:
+        raise ValueError("Cannot use empty string")
+    if node[-1] in ("'", "H"):
+        return HardNode(int(node[:-1]))
+    else:
+        return SoftNode(int(node))
+
+
+def derive_child_key(
```
kclowes: I think it makes sense to group all of these functions that are just floating around in here (`derive_child_key`, `decode_node`, `ec_point`, and  maybe `hmac_sha512`) inside of a class. Maybe call it `HDWallet` or something (but I can't really think of a good name). What do you think?

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,217 @@
...
+      - If not (normal child):
+        let I = HMAC-SHA512(Key = c_par, Data = ser_P(point(k_par)) || ser_32(i)).
+    - Split I into two 32-byte sequences, I_L and I_R.
+    - The returned child key k_i is parse_256(I_L) + k_par (mod n).
+    - The returned chain code c_i is I_R.
+    - In case parse_256(I_L) ≥ n or k_i = 0, the resulting key is invalid,
+      and one should proceed with the next value for i.
+      (Note: this has probability lower than 1 in 2**127.)
+    """
+    assert len(parent_chain_code) == 32
```
kclowes: Instead of `assert`ing throughout this function, I think it would be good to raise more specific errors here with nicer error messages. What do you think?

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __add__(self, other: int):
+        return self.__class__(self.index + other)
+
+    def serialize(self) -> bytes:
+        assert 0 <= self < 2**32
+        return self.to_bytes(4, byteorder="big")
+
+    def encode(self):
+        return str(self.index) + "'"
```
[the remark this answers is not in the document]
fubuloubu: It's a convention. Perhaps I should use `H` here though, so it's clearer most of the time (but still allow `'` to be used for input)

eth_account/hdaccount/deterministic.py
fubuloubu: I was thinking of making `decode_node` a staticmethod of a `Node` base class (per your suggestion about), and making `ec_point` and `hmac_sha512` into functions in a `_utils.py` file under this module.

`derive_child_key` may eventually be accessible to the outside world though, as it may be useful for #25 & #37

eth_account/hdaccount/deterministic.py
fubuloubu: Anywhere you see an `assert` I just wanted to be extra careful to make sure the sizes were correct. They should never be hit in practice, but would cause subtle errors if someone modified the code incorrectly.

eth_account/hdaccount/deterministic.py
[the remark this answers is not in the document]
kclowes: That sounds good to me! Might be worth a comment, too

eth_account/hdaccount/deterministic.py
kclowes: 👍 Yeah, like the Node base class and putting ec_point and hmac_sha512 into a _utils file.

fubuloubu:
@kclowes Made updates per your suggestions!

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+
+    if to_int(child[:32]) >= SECP256K1_N:
+        # Invalid key, compute using next node (< 2**-127 probability)
+        return derive_child_key(parent_key, parent_chain_code, node + 1)
+
+    child_key = (to_int(child[:32]) + to_int(parent_key)) % SECP256K1_N
+    if child_key == 0:
+        # Invalid key, compute using next node (< 2**-127 probability)
+        return derive_child_key(parent_key, parent_chain_code, node + 1)
+    return child_key.to_bytes(32, byteorder="big"), child[32:]
```
kclowes: This is the child private key and the chain code, right?

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class HDPath:
+    def __init__(self, path: str):
+        """
+        Constructor for this class. Initializes an hd account generator and
+        if possible the encoded key and the derivation path. If no arguments are
+        specified, create a new account with createAccount(...) or initialize
+        an account given a mnemonic and an optional password with initAccount(...)
```
kclowes: I don't think this comment is quite accurate, is it coming with the next PR?

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+      (Note: this has probability lower than 1 in 2**127.)
+    """
+    assert len(parent_chain_code) == 32
+    if isinstance(node, HardNode):
+        # NOTE Empty byte is added to align to SoftNode case
+        assert len(parent_key) == 32  # Should be guarenteed here in return statment
+        child = hmac_sha512(parent_chain_code, b"\x00" + parent_key + node.serialize())
+
+    else:
+        assert len(ec_point(parent_key)) == 33  # Should be guarenteed by Account class
```
kclowes: same here: *guaranteed :)

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+* ser_P(P): serializes the coordinate pair P = (x,y) as a byte sequence using SEC1's compressed
+  form: (0x02 or 0x03) || ser_256(x), where the header byte depends on the parity of the
+  omitted y coordinate.
+* parse_256(p): interprets a 32-byte sequence as a 256-bit number, most significant byte first.
+
+"""
+# Additional notes:
+# - This algorithm only implements private parent key → private child key CKD function,
+#   as it is unnecessary to the HD key derivation functions used in this library to implement
+#   the other functions (as Ethereum uses an Account-based system)
```
kclowes: This is 100% my own ignorance, but why does Ethereum using an Account-based system impact why we don't need the other CKD functions? It seems like we would need at least public -> public?

fubuloubu:
@kclowes that's correct. The seed gets generated using BIP39

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+        specified, create a new account with createAccount(...) or initialize
+        an account given a mnemonic and an optional password with initAccount(...)
+        :param path             : BIP32-compatible derivation path
+        :type path              : str as "m/idx_0/.../idx_n" or "m/idx_0/.../idx_n"
+                                  where idx_* is either an integer value (soft node)
+                                  or an integer value followed by either the "'" char
+                                  or the "H" char (hardened node)
+        """
+        nodes = path.split('/')
+        if not nodes[0] == 'm':
```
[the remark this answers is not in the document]
fubuloubu: I don't think it's part of the spec, but shouldn't harm things I wouldn't think?

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def encode(self) -> str:
+        encoded_path = ['m']
+        for node in self._path:
+            encoded_path.append(node.encode())
+        return '/'.join(encoded_path)
+
+    def derive(self, seed: bytes) -> bytes:
+        master_node = hmac_sha512(b"Bitcoin seed", seed)
+        key = master_node[:32]
```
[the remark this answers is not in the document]
fubuloubu: I'm trying to start the recursion here with this value.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+    - The returned child key k_i is parse_256(I_L) + k_par (mod n).
+    - The returned chain code c_i is I_R.
+    - In case parse_256(I_L) ≥ n or k_i = 0, the resulting key is invalid,
+      and one should proceed with the next value for i.
+      (Note: this has probability lower than 1 in 2**127.)
+    """
+    assert len(parent_chain_code) == 32
+    if isinstance(node, HardNode):
+        # NOTE Empty byte is added to align to SoftNode case
+        assert len(parent_key) == 32  # Should be guarenteed here in return statment
```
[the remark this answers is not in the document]
fubuloubu:
```suggestion
        assert len(parent_key) == 32  # Should be guaranteed here in return statment
```

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+            for node in nodes[1:]:  # We don't need the root node 'm'
+                decoded_path.append(Node.decode(node))
+        except ValueError as e:
+            raise ValueError(f'Path is not valid: "{path}". Issue with node "{node}": {e}')
+        self._path = decoded_path
+
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}(path="{self.encode()}")'
+
+    def encode(self) -> str:
```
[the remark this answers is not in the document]
fubuloubu: Nope, just for `__repr__`, for pretty-printing!

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,192 @@
...
+) -> Tuple[bytes, bytes]:
+    """
+    From BIP32:
+
+    The function CKDpriv((k_par, c_par), i) → (k_i, c_i) computes a child extended
+    private key from the parent extended private key:
+
+    - Check whether i ≥ 2**31 (whether the child is a hardened key).
+      - If so (hardened child):
+        let I = HMAC-SHA512(Key = c_par, Data = 0x00 || ser_256(kpar) || ser_32(i)).
```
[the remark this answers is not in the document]
fubuloubu:
```suggestion
        let I = HMAC-SHA512(Key = c_par, Data = 0x00 || ser_256(k_par) || ser_32(i)).
```

eth_account/hdaccount/deterministic.py
fubuloubu: Yeah, I could write it out on the line before to make it more explicit

eth_account/hdaccount/deterministic.py
fubuloubu: Oh yes, that comment is not accurate! Copy paste errors :grimacing:

eth_account/hdaccount/deterministic.py
fubuloubu:
```suggestion
        assert len(ec_point(parent_key)) == 33  # Should be guaranteed by Account class
```

eth_account/hdaccount/deterministic.py
fubuloubu: Your guess is as good as mine for why people decided to take this path. Probably to be interoperable with Hardware wallets, which are built using HD Path derivation (for multiple coins)

fubuloubu: Actually, this comment is wrong. Account-based system doesn't really matter, I think I was trying to explain that the use of change addresses and continuous key generation (as a UTXO model would use to prevent key reuse) isn't a necessary factor here. But that seems like unnecessary information.

eth_account/hdaccount/deterministic.py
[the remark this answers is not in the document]
kclowes: Cool. Yeah, I didn't see it as part of the spec either but saw that py-hdwallet allowed it. I'll leave that decision up to you!

fubuloubu:
Any updates on merging this PR?

pipermerriam:
Lemme give it a pass and I'll leave the final merge up to @kclowes

eth_account/hdaccount/_utils.py
```diff
@@ -0,0 +1,27 @@
+import hashlib
```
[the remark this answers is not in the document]
pipermerriam: There's an established precedent for doing it this way, and since we plan to yank this out in favor of py-hdwallet in the near term future, I think it makes the most sense to leave it here since it will make that extraction easier.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,211 @@
...
+        return obj
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.index})"
+
+    def __add__(self, other: int):
+        return self.__class__(self.index + other)
+
+    def serialize(self) -> bytes:
+        assert 0 <= self < 2**32
```
pipermerriam: What do you think about moving this validation to `__new__` so that we can prevent an invalid one of these from even being instantiated?

Also, we generally stay away from `assert thing` in runtime code because of [how easy it is to disable them at runtime](https://stackoverflow.com/questions/1273211/disable-assertions-in-python).  Using `ValueError` is probably acceptable here if you want to stick with standard library, or `eth_utils.ValidationError` is a common convention across our libraries as well.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,211 @@
...
+      and one should proceed with the next value for i.
+      (Note: this has probability lower than 1 in 2**127.)
+    """
+    assert len(parent_chain_code) == 32
+    if isinstance(node, HardNode):
+        # NOTE Empty byte is added to align to SoftNode case
+        assert len(parent_key) == 32  # Should be guaranteed here in return statment
+        child = hmac_sha512(parent_chain_code, b"\x00" + parent_key + node.serialize())
+
+    else:
```
pipermerriam: We have trended away from using *implicit* else clauses.  Am I correct that this case covers the `SoftNode` case?  If so I would advocate for an explicit check for `elif isinstance(node, SoftNode)` and then to have a final `else` clause that raises an exception like `TypeError` indicating that the `node` was not of a supported type.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,211 @@
...
+        for examples of how this object may be used. Please note however that this
+        object makes no such assumptions of the use of BIP43 or BIP44, or later BIPs.
+        :param path             : BIP32-compatible derivation path
+        :type path              : str as "m/idx_0/.../idx_n" or "m/idx_0/.../idx_n"
+                                  where idx_* is either an integer value (soft node)
+                                  or an integer value followed by either the "'" char
+                                  or the "H" char (hardened node)
+        """
+        nodes = path.split('/')
+        if nodes[0] not in ('m', 'M'):
```
pipermerriam: If we want to be really careful we should pre-validate that `path` is not the empty string, otherwise this will raise an `IndexError` since `nodes` will be the empty list.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,211 @@
...
+            raise ValueError(f'Path is not valid: "{path}". Must start with "m"')
```
pipermerriam: you could do this validation before splitting the path which would prevent the `IndexError` case I mentioned above.

eth_account/hdaccount/deterministic.py
pipermerriam: Also, can you move `('m', 'M')` to be a module level constant and make it a `set` since it is just used for inclusion checks.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,211 @@
...
+        decoded_path = []
+        try:
+            for node in nodes[1:]:  # We don't need the root node 'm'
+                decoded_path.append(Node.decode(node))
+        except ValueError as e:
+            raise ValueError(f'Path is not valid: "{path}". Issue with node "{node}": {e}')
```
pipermerriam: It would come at a slight performance cost but if you move the `try/except` inside the `for` loop, and change the `for` loop to use `enumerate` you can improve this error message by including the index of the node.

```python
for idx, node in enumerate(nodes[1:]):
    try:
        ...
    except ValueError as err:
        raise ValueError(...) from err
```

Also, using the `raise ... from ...` syntax *can* be nice as it preserves the entirety of the stacktrace.  Maybe not beneficial in this context but it can be nice when you want to know where the inner error came from.

eth_account/hdaccount/deterministic.py
```diff
@@ -0,0 +1,211 @@
...
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}(path="{self.encode()}")'
+
+    def encode(self) -> str:
+        """
+        Encodes this class to a string (reversing the decoding in the constructor)
+        """
+        encoded_path = ['m']
+        for node in self._path:
+            encoded_path.append(node.encode())
```
pipermerriam: *mutability-nitpick*

Would you be ok modeling this as.  

```python
encoded_path = ('m',) + tuple(node.encode() for node in self._path)
```

Both personal preference, and wanting to avoid enshrining mutability patterns into the code for fear that it encourages more of it.  Immutable by default unless we have an explicit reason to need it is the general policy.

eth_account/hdaccount/deterministic.py
fubuloubu: I'll fix this one up, I did `assert` elsewhere just as an extra size check, even though badly-sized input shouldn't be used outside of this library

eth_account/hdaccount/deterministic.py
fubuloubu: Will do

eth_account/hdaccount/deterministic.py
fubuloubu: Will do

eth_account/hdaccount/deterministic.py
fubuloubu: Will do

eth_account/hdaccount/deterministic.py
fubuloubu: Will do

eth_account/hdaccount/deterministic.py
fubuloubu: Will do

kclowes:
@fubuloubu I'll merge this shortly! I started going through BIP39 today and should have a review for that PR soon.

As general feedback on the HD wallet feature: I spoke with @pipermerriam and before we release it, we'll want to put it behind a flag that makes this API inaccessible by default because it hasn't been audited. Here is an example of putting EthPM behind a similar sort of flag in web3: https://github.com/ethereum/web3.py/pull/1216. I _think_ we can get away with only adding the flag in eth-account, but we may also want to add it to web3.

And when you're adding documentation, we'll also need to add a warning there so that users know to use at their own risk.

fubuloubu:
@kclowes sounds good, however I think we can talk about this when we get to #87, as that's where all the API stuff is

kclowes:
Yep, that sounds good to me!

Issue #90, opened 2020-02-05
Cannot encode bytes32 in structured data


jstoxrocky:
## What was wrong?
I cannot encode `bytes32` in structured data. I am attempting to pass a struct containing a variable I want encoded as `bytes32` to `eth_account.messages.encode_structured_data`. I am passing this struct to `encode_structured_data` as a Python dictionary. Including my `bytes32` variable as a bytestring causes a JSON serialization error (eth_account/messages.py:143). Passing this variable as a hex string throws a TypeError (eth_account/_utils/structured_data/hashing.py:250).

### Code that produced the error

```py
from eth_account.messages import (
    encode_structured_data,
)
from eth_utils import (
    keccak,
)

message = {
	"types": {
		"EIP712Domain": [
			{"name": "name", "type": "string"},
			{"name": "version", "type": "string"},
			{"name": "chainId", "type": "uint256"},
			{"name": "verifyingContract", "type": "address"}
		],
		"coolStruct": [
			{"name": "coolBytes", "type": "bytes32"}
		]
	},
	"primaryType": "coolStruct",
	"domain": {
		"name": "coolContract",
		"version": "1",
		"chainId": 1,
		"verifyingContract": "0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC"
	},
	"message": {
		"coolBytes": keccak(b''), # Alternativley keccak(b'').hex()
	}
}

def test_hashed_structured_data():
    structured_msg = encode_structured_data(message)
```

### Full error output 

```sh
eth_account/messages.py:143: in encode_structured_data
    message_string = json.dumps(primitive)
TypeError: Object of type bytes is not JSON serializable
```

```sh
TypeError: Value of `coolBytes` (c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) in the struct `coolStruct` is of the type `<class 'str'>`, but expected bytes32 value
```

### Expected Result
It can definitely be done and should equal something like this:
```py
    domain_type_hash = keccak(text='EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)')
    name = message['domain']['name']
    version = message['domain']['version']
    chain_id = message['domain']['chainId']
    verifying_contract = message['domain']['verifyingContract']
    domain_instance = encode_abi(
      ['bytes32', 'bytes32', 'bytes32', 'uint256', 'address'],
[...]

PR #91, opened 2020-02-05
Allow encoding of structured data containing bytes32

jstoxrocky:
## What was wrong?
The json serialization in eth_account/messages.py:143 did not allow for bytes data to be used in a struct to be signed.

Issue #90 

## How was it fixed?
If the struct is a Mapping then we can skip the `json.dumps` -> `json.loads` pattern.

#### Cute Animal Picture

axolotl

### DIFF
--- a/eth_account/messages.py
+++ b/eth_account/messages.py
@@ -1,7 +1,6 @@
 from collections.abc import (
     Mapping,
 )
-import json
 from typing import (
     NamedTuple,
     Union,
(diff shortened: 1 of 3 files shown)


fselmo (approved):
This looks good to me. We still need to validate types in the structured data but I created #140 to track this as it is outside of this scope. I'm going to add a newsfragment and merge. Thanks again!

PR #94, opened 2020-03-31
Doctests for Account and other general doc cleanup

kclowes:
## What was wrong?
There were no doctests in the eth-account docs.

Issue #9 

## How was it fixed?
Added some!

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/signers/local.py
+++ b/eth_account/signers/local.py
@@ -46,7 +46,7 @@ class LocalAccount(BaseAccount):
     @property
     def privateKey(self):
         """
-        .. CAUTION:: Deprecated for :var:`~eth_account.signers.local.LocalAccount.key`.
+        .. CAUTION:: Deprecated for :meth:`~eth_account.signers.local.LocalAccount.key`.
             This attribute will be removed in v0.5
         """
         warnings.warn(
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,6 +82,7 @@ exclude_patterns = [
     '_build',
     'modules.rst',
     'eth_account.internal.rst',
+    'eth_account.hdaccount*',
 ]
 
 # The reST default role (used for this markup: `text`) to use for all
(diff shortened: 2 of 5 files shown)


eth_account/signers/local.py
```diff
@@ -46,7 +46,7 @@ def address(self):
     @property
     def privateKey(self):
         """
-        .. CAUTION:: Deprecated for :var:`~eth_account.signers.local.LocalAccount.key`.
+        .. CAUTION:: Deprecated for :meth:`~eth_account.signers.local.LocalAccount.key`.
```
carver, line 49: Ah interesting, so the link changes to `:meth:` when it's a `@property`. Did this show up during doc tests? You were able to confirm that it links correctly with this change?

kclowes: Yep, it was showing up as a warning in the doc build. And it does work correctly!

kclowes:
@carver I had run the tests using the circleci CLI locally, but not all together, so I'm glad I did that. Good suggestion! It looks like the integration and core tests were running tests in the `eth_account` directory as well, so I took those out from the `tox.ini` (see the [last commit](https://github.com/ethereum/eth-account/pull/94/commits/808127f5428ca5f1b73039056ef04b9d7342ac8e)), but let me know if that doesn't sound right to you.

carver:
I had a vague memory that they were intentional. Luckily it doesn't change too often, so a git blame quickly showed the commit that added it: https://github.com/ethereum/eth-account/commit/8d99dea671a89a4c213abe4745e331830c35b386

It does look like it's necessary to leave it in so that doctest runs against any docstring examples. I think the inclusion in the integration test is just copy/pasta though:
https://github.com/ethereum/eth-account/commit/780a7cf067827742a06ef0968fd661e2f48f5a8f#diff-b91f3d5bd63fcd17221b267e851608e8

Although, one option would be to leave the change you did already and then add an extra tox.ini command under the `doctest` environment: `pytest eth_account`.

That way any docstring doctest failures would show up in the `doctest` CI job instead of the `core` job. That's probably more intuitive.

PR #96, opened 2020-04-07
wip: introduce py-hdwallet

wolovim:
## What does it do?

Introduces [py-hdwallet](https://github.com/ethereum/py-hdwallet) for child key derivation.

(WIP: seeking feedback on items below)

## Background

- My understanding of __the plan__ is for py-hdwallet to eventually provide all BIP32-related functionality within eth-account. This introduction of the library only goes as far as replicating existing functionality recently introduced by @fubuloubu.
- Mnemonics are currently outside of the scope of py-hdwallet, so @fubuloubu's recent mnemonic work would go untouched.

### TODO:
- [ ] put all testing in its right place, e.g., remove or alter relevant tests in eth-account; some tests are better suited for py-hdwallet.
- [ ] agree on hardening representation (e.g., `'`, `h`, `H`, etc.)
   - py-hdwallet only supports `h`
   - the eth-account implementation had support for `'` and `H`

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
         "eth-rlp>=0.1.2,<1",
         "eth-utils>=1.3.0,<2",
         "hexbytes>=0.1.0,<1",
+        "py-hdwallet>=0.1.0a1",
         "rlp>=1.0.0,<2"
     ],
     setup_requires=['setuptools-markdown'],
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -49,7 +49,6 @@ from eth_account.datastructures import (
 from eth_account.hdaccount import (
     ETHEREUM_DEFAULT_PATH,
     generate_mnemonic,
-    key_from_seed,
     seed_from_mnemonic,
 )
 from eth_account.messages import (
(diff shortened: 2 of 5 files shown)


fubuloubu:
The `'` is the standard representation IIRC. It should be supported prior to either of the other ones. Understanding it's not exactly the most legible in code, but a simple replace would solve that.

fubuloubu:
Also, a nice thing that should be introduced at some point is an alternative account type that can continuously derive new accounts using hdwallet, and encrypts/decrypts the master key instead of the account key. This would come in handy primarily for testing use (deriving multiple test accounts from a seed), but would set the stage for hardware wallet integration (which uses hd paths to obtain the account from master public key).

setup.py
```diff
@@ -56,6 +56,7 @@
         "eth-rlp>=0.1.2,<1",
         "eth-utils>=1.3.0,<2",
         "hexbytes>=0.1.0,<1",
+        "py-hdwallet>=0.1.0a1",
```
pipermerriam, line 59: BTW, we'd typically want to at least put `py-hdwallet` into beta before releasing this.

stomde:
Merge me plz.... luv

PR #97, opened 2020-04-21
Fix crash in structured messages with arrays

carver:
## What was wrong?

Arrays in structured messages were badly broken, raising a `RecursionError`. I guess no one has tried it yet (or they haven't reported it, that I've seen).

## How was it fixed?

- mypy highlighted the problem
- structured types with arrays were not tested, so add an array type to the existing test
- fix a couple issues:
  - arrays could not be flattened properly, blowing up the stack on infinite
recursion
  - a list of bytes was trying to be joined with a str

The array flattening seemed to have some backward logic, flipped it.
Switched the element merging to join with a `bytes` value.

#### Cute Animal Picture

### DIFF
--- a/eth_account/_utils/structured_data/hashing.py
+++ b/eth_account/_utils/structured_data/hashing.py
@@ -156,7 +156,7 @@ def get_array_dimensions(data):
 @to_tuple
 def flatten_multidimensional_array(array):
     for item in array:
-        if not isinstance(item, (list, tuple)):
+        if isinstance(item, (list, tuple)):
             # Not checking for Iterable instance, because even Dictionaries and strings
             # are considered as iterables, but that's not what we want the condition to be.
             yield from flatten_multidimensional_array(item)
@@ -229,7 +229,7 @@ def _encode_data(primary_type, types, data):
                 encode_data(parsed_type.base, types, array_item)
                 for array_item in array_items
             ]
-            concatenated_array_encodings = ''.join(array_items_encoding)
+            concatenated_array_encodings = b''.join(array_items_encoding)
             hashed_value = keccak(concatenated_array_encodings)
             yield "bytes32", hashed_value
         else:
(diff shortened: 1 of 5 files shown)


eth_account/_utils/structured_data/hashing.py
```diff
@@ -156,7 +156,7 @@ def get_array_dimensions(data):
 @to_tuple
 def flatten_multidimensional_array(array):
     for item in array:
-        if not isinstance(item, (list, tuple)):
+        if isinstance(item, (list, tuple)):
```
carver, line 159: @Bhargavasomu I *think* this was just a flipped logic bug. Does anything here look wrong to you?

tests/fixtures/valid_message.json
```diff
@@ -32,6 +33,16 @@
 			"name": "Bob",
 			"wallet": "0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB"
 		},
+		"cc": [
+      {
+        "name": "Alice",
+        "wallet": "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
carver, line 39: Hm, this probably should have triggered a checksum error. I wonder why it didn't?

pipermerriam:
ping @marcgarreau as if you don't have enough new things flying at you, this is an area it wouldn't hurt to get more familiar with and it falls pretty smack-dab into the end-user-APIs category.

eth_account/_utils/structured_data/hashing.py
Bhargavasomu: Hi @carver, it's been a long time since I wrote that code, and I do believe it is a flipped logic bug. Thank you for catching it. As far as the function definition and the usage goes, I don't see anything wrong with this change.

tests/fixtures/valid_message.json
carver: I started down the path of enforcing checksum. To my great horror, the checksum version of the address did not encode the same way, it changed the hash!

So I settled for increasing the warning in the docs about the lack of tests.

carver:
With a fairly aggressive warning about this not being well-tested (added to the docs), I'll go ahead and merge this bugfix.

PR #100, opened 2020-04-22
WIP: Require checksum addresses in typed/structured messages

carver:
## What was wrong?

Checksummed addresses weren't enforced. Seems like something worth doing.

## How was it fixed?

I tried adding a checksum requirement. Then when I changed the addresses to be checksummed... the hash changed! :scream: (whether the type was left as "address" or "uint160")

I don't have time to dig in at this moment, so this is just a starting point for someone else or me to pick up sometime later.

#### Cute Animal Picture

### DIFF
--- a/eth_account/_utils/structured_data/hashing.py
+++ b/eth_account/_utils/structured_data/hashing.py
@@ -16,7 +16,10 @@ from eth_abi.grammar import (
 )
 from eth_utils import (
     ValidationError,
+    is_checksum_address,
+    is_checksum_formatted_address,
     keccak,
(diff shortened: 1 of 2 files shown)


pacrob:
We're moving towards being less restrictive in typed data signing, given that a message may originate from a variety of places and we don't want users to have to do a bunch of cleanup to get it signed. I checked against current versions of ethers and eth-sig-util (metamask) and neither enforce checksummed addresses. They do, as does our current `encode_structured_data`, turn out the same signature whether addresses in the message are checksummed or not.

PR #108, opened 2020-10-08
Relax upper bound of dependency

cburgdorf:
### What was wrong?

The current RLP dependency has a hard bound on `2.0.0a1`. Trinity, Py-EVM and friends need to migrate to `2.0.0a2` so we need to update the dependency. 

### How was it fixed?

Changed it to `"rlp>=1,<3"` which is  inline with what [`eth-enr` does](https://github.com/ethereum/eth-enr/blob/master/setup.py#L64) and causes less cascading releases whenever we cut a new alpha release of RLP.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,10 +61,10 @@ setup(
         "eth-abi>=2.0.0b7,<3",
         "eth-keyfile>=0.5.0,<0.6.0",
         "eth-keys>=0.2.1,<0.4.0,!=0.3.2",
-        "eth-rlp>=0.1.2,<1",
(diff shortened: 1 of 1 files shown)


pipermerriam (approved):
same question as https://github.com/ethereum/eth-tester/pull/201#pullrequestreview-507457016

cburgdorf:
@pipermerriam @carver I need someone to either:

- merge and release
- give me rights for GitHub and PyPi

PR #111, opened 2021-03-19
feat: add EIP712 message subclasses

fubuloubu:
Added an implementation of EIP712 structured data EIP as a hierarchy of subclasses that users can define according to their needs.

### DIFF
--- a/eth_account/_utils/structured_data/validation.py
+++ b/eth_account/_utils/structured_data/validation.py
@@ -63,21 +63,30 @@ def validate_field_declared_only_once_in_struct(field_name, struct_data, struct_
         )
 
 
+EIP712_DOMAIN_FIELDS = [
+    "name",
+    "version",
+    "chainId",
+    "verifyingContract",
+]
+
+
+def used_header_fields(EIP712Domain_data):
+    return [field["name"] for field in EIP712Domain_data if field["name"] in EIP712_DOMAIN_FIELDS]
+
+
 def validate_EIP712Domain_schema(structured_data):
     # Check that the `types` attribute contains `EIP712Domain` schema declaration
     if "EIP712Domain" not in structured_data["types"]:
         raise ValidationError("`EIP712Domain struct` not found in types attribute")
     # Check that the names and types in `EIP712Domain` are what are mentioned in the EIP-712
-    # and they are declared only once
+    # and they are declared only once (if defined at all)
     EIP712Domain_data = structured_data["types"]["EIP712Domain"]
-    validate_field_declared_only_once_in_struct("name", EIP712Domain_data, "EIP712Domain")
-    validate_field_declared_only_once_in_struct("version", EIP712Domain_data, "EIP712Domain")
-    validate_field_declared_only_once_in_struct("chainId", EIP712Domain_data, "EIP712Domain")
-    validate_field_declared_only_once_in_struct(
-        "verifyingContract",
-        EIP712Domain_data,
-        "EIP712Domain",
-    )
+    header_fields = used_header_fields(EIP712Domain_data)
+    if len(header_fields) == 0:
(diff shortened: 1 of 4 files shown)


eth_account/messages.py
```diff
@@ -56,6 +62,102 @@ class SignableMessage(NamedTuple):
...
+                types[self.type].append({"name": field, "type": self.field_type(field)})
+
+        return types
+
+    @property
+    def data(self) -> dict:
+        return as_dict(self)  # NOTE: Handles recursion
+
+
+# TODO: Make type of EIP712Message a subtype of SignableMessage somehow
```
pipermerriam, line 110: If this proves *difficult* maybe just a convenience API like `MyEIP712Message.as_signable_message()`

carver: In fact, whether or not it is difficult to subclass, I'd consider the "emit signable message" approach to be preferable.

fubuloubu: It is more explicit

eth_account/_utils/structured_data/validation.py
```diff
@@ -63,21 +63,30 @@ def validate_field_declared_only_once_in_struct(field_name, struct_data, struct_
...
-    validate_field_declared_only_once_in_struct(
-        "verifyingContract",
-        EIP712Domain_data,
-        "EIP712Domain",
-    )
+    header_fields = used_header_fields(EIP712Domain_data)
+    if len(header_fields) == 0:
+        raise ValidationError(f"One of {EIP712_DOMAIN_FIELDS} must be defined in {structured_data}")
+    for field in header_fields:
+        validate_field_declared_only_once_in_struct(field, EIP712Domain_data, "EIP712Domain")
```
fubuloubu, line 89: I might move this change to it's own PR, it fixes a bug in how this library currently handles EIP712 Domain validation

eth_account/messages.py
pipermerriam: I think the intent is to keep the signing API dead simple, and to have `EIP712Message.as_signable_message()` emit that simpler `SignableMessage` API.

carver: Actually, all of this about creating a signed message seems like the wrong direction. The pattern in eth-account is to use the signable-message-building methods. In fact there is already one for 712:
https://github.com/ethereum/eth-account/blob/c4bc769039fdacb8eddf26c84640af84651a9da2/eth_account/messages.py#L114

I'm signing off soon, but I'll come back and take a look to see if we can find another approach. Duplicate ways of building 712 signatures doesn't seem like the right approach, right now.

fubuloubu: Could turn this into another library, just not really a huge fan of how the current API works because it really doesn't fit well with how I'd want to use it, which is more like creating a struct and leveraging that several times. This approach using dataclasses seemed much cleaner from an end user perspective, but I could definitely see how it might better live in another library or `web3py`

eth_account/_utils/structured_data/validation.py
fubuloubu: See #112

eth_account/messages.py
carver: Yeah, I think it's very reasonable for it to live in its own library. As long as the library can generate a `SignableMessage`, then it should be trivial to just use the existing eth-account APIs. It doesn't really need to live in here.

fubuloubu: Will look into this, when I have more time.

carver: Cool. Once a separate 712 implementation looks good, I think we would probably just strip out the one already in here.

fubuloubu: Yeah this one is pretty awkward to use right now

fubuloubu:
Closing with work from #112 and discussion from above

PR #127, opened 2021-09-21
doc: replace dup method with missing method

unparalleled-js:
## What was wrong?

A factory-like method was listed twice in the `SignableMessage` documentation and one of the factory-like methods was missing  

## How was it fixed?

Replaced duplicated method with the missing one in the doc

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

🐕


kclowes (approved):
Awesome, thank you! :rocket: I'll add a newsfragment and merge here too!

PR #128, opened 2021-09-21
fix: add missing f-str

unparalleled-js:
## What was wrong?

Issue # One of the validation error was missing an `f` in its string-interpolation 

## How was it fixed?

Prefixed str with `f`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

🐶


kclowes (approved):
Thank you! I'll add a newsfragment and merge!

PR #138, opened 2022-01-19
Update eth-abi, eth-rlp, eth-keyfile, eth-keys, pyrlp, eth-utils

kclowes:
## What was wrong?

Need to update all these dependencies to keep everything copacetic. This should be merged after #139 so tests pass.

## How was it fixed?

Updated `eth-abi`, `eth-rlp`, `eth-keyfile`, `eth-keys`, `pyrlp`, and `eth-utils` dependencies. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/test_structured_data_signing.py
+++ b/tests/core/test_structured_data_signing.py
@@ -2,9 +2,6 @@ import json
 import pytest
 import re
 
-from eth_abi.exceptions import (
-    ABITypeError,
-)
 from eth_utils import (
     ValidationError,
     keccak,
@@ -223,12 +220,12 @@ def test_invalid_structured_data_value_type_mismatch_in_primary_type():
         "tests/fixtures/invalid_message_value_type_mismatch_primary_type.json"
     ).read()
     invalid_structured_data = json.loads(invalid_structured_data_string)
-    with pytest.raises(TypeError) as e:
+
+    with pytest.raises(
+        TypeError,
+        match="in the struct `Mail` is of the type `<class 'int'>`, but expected string value"
+    ):
         hash_message(invalid_structured_data)
-    assert (
-        str(e.value) == "Value of `contents` (12345) in the struct `Mail` is of the "
-        "type `<class 'int'>`, but expected string value"
-    )
 
 
 def test_invalid_structured_data_invalid_abi_type():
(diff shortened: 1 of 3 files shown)


tests/core/test_structured_data_signing.py
```diff
@@ -237,9 +234,8 @@ def test_invalid_structured_data_invalid_abi_type():
         "tests/fixtures/invalid_message_invalid_abi_type.json"
     ).read()
     invalid_structured_data = json.loads(invalid_structured_data_string)
-    with pytest.raises(ABITypeError) as e:
+    with pytest.raises(TypeError, match="Received Invalid type `uint25689` in the struct `Person`"):
```
kclowes, line 237: This is a small breaking change, but decided to go for it since this release will be breaking with the dependency updates anyways. `eth-abi` had a bug where `ABITypeError` was raised on a `has_encoder` check instead of returning `False`. That got fixed downstream in `eth-abi`, which caused this test to fail. Raising this `TypeError` here is consistent with other validation checks for signing structured data.

fselmo: Yeah makes sense... I'm good with it 👍

fselmo (approved):
LGTM 👍

kclowes:
Pulling this into web3.py will be a breaking change for web3.py, so we'll have to decide how we want to handle that. The options I can think of are:

1) Put up a v6 branch for web3 now-ish, but would still be unstable for a while yet until breaking changes are in
2) Sit on this for a while longer and don't pull into web3.py (but we may run into problems with eth-tester)
3) Go back through and make a new minor release for each library that keeps all old python versions, and upgrade when we're ready to start the v6 branch

I think I lean toward 1 or 3, but let me know if anyone has preferences or if you can think of a better solution!

fselmo:
> Pulling this into web3.py will be a breaking change for web3.py, so we'll have to decide how we want to handle that. The options I can think of are:
> 
>     1. Put up a v6 branch for web3 now-ish, but would still be unstable for a while yet until breaking changes are in
>     2. Sit on this for a while longer and don't pull into web3.py (but we may run into problems with eth-tester)
>     3. Go back through and make a new minor release for each library that keeps all old python versions, and upgrade when we're ready to start the v6 branch

I lean towards 1 here. This would give us the most flexibility in merging other breaking issues as well. Not against 3 if there are good enough reasons to wait on a v6 branch though.

kclowes:
I'm good with that. I think that makes the most sense. If it turns out to be a problem, we can always go back and do a minor for the libraries that need it. 🎉 v6 here we come! 🚂

fselmo:
> I'm good with that. I think that makes the most sense. If it turns out to be a problem, we can always go back and do a minor for the libraries that need it. 🎉 v6 here we come! 🚂

💥  _itshappening.gif_

PR #139, opened 2022-01-19
Drop Python 3.6 support, Add Python 3.9-3.10

kclowes:
## What was wrong?

Dependencies have dropped python 3.6 support, so that needs to be dropped here too to be compatible.

## How was it fixed?

Dropped python 3.6 CI integration, and added tests for Python 3.9 and 3.10

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,21 +79,15 @@ jobs:
   docs:
     <<: *common
     docker:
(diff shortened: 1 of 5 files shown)


pacrob (approved):
lgtm!

PR #142, opened 2022-01-27
Bump eth keys

kclowes:
## What was wrong?

Needed to pull in the ecrecover point at infinity bugfix from eth-keys.

## How was it fixed?

`eth-keys` v0.3.4 adds the fix, so bumped the dependency requirement and added a test here. I couldn't decide if we should catch the `eth-keys` exception `BadSignature` and re-raise with a different error/message or if we should just leave it. I opted to just leave it for now, but happy to catch if someone feels strongly.

This is PR-ed against a v0.5.x branch that contains the non-breaking changes, so I'll merge this in and release 0.5.7 once this is approved. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,6 @@ crashlytics-build.properties
 fabric.properties
 
 # END JetBrains section
+
+# Jupyter notebook files
+*.ipynb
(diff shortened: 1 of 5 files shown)


pacrob (approved):
looks good, BadSignature seems clear enough to me

fselmo (approved):
No strong opinion on the exception + message here, lgtm 👍

PR #143, opened 2022-02-01
chore: update eth-abi dependency

fubuloubu:
## What was wrong?
`eth-account` had one old pin on `eth-abi` that was missed with recent updates.

## How was it fixed?
Update pin

#### Cute Animal Picture
cute frogs


fubuloubu:
whoops, I totally read this wrong lol

Issue #140, opened 2022-01-21
Validate types in `encode_structured_data()`


fselmo:
## What was wrong?

We currently don't seem to validate types when using `encode_structured_data` (i.e. `bytes32`, `uint8`, `uint256` sizes).

_Relevant issues that sparked the creation of this issue:_ #90, #91

## How can it be fixed?

- We should add type-specific validation to ensure that the values for these fields do not exceed the sizes they are defined by
- Investigate if this validation should be added anywhere else in the code base

fselmo:
This was unlikely to begin with... it turns out we validate the values when hashing, not during the "validation" steps. Those seem stick to validating that supported type values, as strings, as provided in the fields. However, errors stemming from passing in the wrong types could be improved. The messaging seems to be good for when the base types between the provided `type` differs from the actual `value` but not when the base type is the same but exceeds the provided size.

e.g. If the type is `uint8` and the value is `256`... the error will print out that we expected `uint8` but got type `int`. In reality, the value being an `int` is fine, it just exceeds the size allowed for encoding a `uint8` type.

This is a bit minor but would improve the user experience. However, it can be addressed in a separate issue. Closing this since we do validate fields appropriately.

PR #144, opened 2022-02-22
Better messaging when provided `value` cannot be encoded as specified `type`

fselmo:
## What was wrong?

- Unclear messaging when the `value` provided for a specified `type` in structured data matches the base type but exceeds the specified size.
- Related message in issue #140

## How was it fixed?

- Clearer message based on the logic that is happening. In the logic surrounding this, we check if the provided `value` can be encoded as the provided `type` via `is_encodable()`. The messaging was changed to be a bit more clear that the `value` provided is not encodable as the specified `type`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_account/_utils/structured_data/hashing.py
+++ b/eth_account/_utils/structured_data/hashing.py
@@ -246,20 +246,15 @@ def _encode_data(primary_type, types, data):
                     )
                 )
 
-            # Next see if the data fits the specified encoding type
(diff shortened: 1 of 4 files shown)


pacrob (approved):
lgtm!

kclowes (approved):
LGTM! :rocket:

PR #145, opened 2022-02-23
Add and update tests for encode_structured_data

### LINKED ISSUE #68: Add tests to show that tuples are handled correctly in typed data signing
I guess #67 implicitly adds support for tuple types in the typed data signing, maybe more that weren't previously supported. Probably worth adding some tests to prove it and make sure we don't have regressions.

fselmo:
## What was wrong?

- I believe this closes #68 since tuples aren't directly passed in to structured data, only structs.

## How was it fixed?

* Add more robust tests for nested tuples / structs for encode structured data (#68)
* Put back tests that directly test the example from EIP-712 so it can be referenced and checked against actual values if broken
* Refactor tests so that most of the structured data / structured data signing tests test both the EIP-712 example and the updated example with an array

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/test_structured_data_signing.py
+++ b/tests/core/test_structured_data_signing.py
@@ -34,40 +34,81 @@ from eth_account.messages import (
 
 
 @pytest.fixture
-def structured_valid_data_json_string():
-    return open("tests/fixtures/valid_message.json", "r").read()
+def eip712_example_json_string():
+    return open("tests/fixtures/valid_eip712_example.json", "r").read()
 
 
 @pytest.fixture
-def types(structured_valid_data_json_string):
-    return json.loads(structured_valid_data_json_string)["types"]
+def eip712_example_types(eip712_example_json_string):
+    return json.loads(eip712_example_json_string)["types"]
 
 
 @pytest.fixture
-def domain_type(structured_valid_data_json_string):
-    return json.loads(structured_valid_data_json_string)["types"]["EIP712Domain"]
+def eip712_example_domain_type(eip712_example_json_string):
+    return json.loads(eip712_example_json_string)["types"]["EIP712Domain"]
 
 
 @pytest.fixture
-def message(structured_valid_data_json_string):
-    return json.loads(structured_valid_data_json_string)["message"]
+def eip712_example_message(eip712_example_json_string):
+    return json.loads(eip712_example_json_string)["message"]
 
 
 @pytest.fixture
-def domain(structured_valid_data_json_string):
-    return json.loads(structured_valid_data_json_string)["domain"]
+def eip712_example_domain(eip712_example_json_string):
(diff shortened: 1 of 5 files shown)


pacrob (approved):
lgtm!

tests/core/test_structured_data_signing.py
```diff
@@ -34,124 +34,263 @@
...
-    expected_hex_value = "f2cee375fa42b42143804025fc449deafd50cc031ca257e0b194a650a912090f"
-    assert hash_domain(structured_data).hex() == expected_hex_value
+@pytest.mark.parametrize('example', ('eip712', 'eip712_with_array',))
+def test_hash_struct_domain(
+    example, eip712_example_json_string, eip712_example_with_array_json_string
+):
+    structured_data = (
+        json.loads(eip712_example_json_string) if example == 'eip712'
+        else json.loads(eip712_example_with_array_json_string) if example == 'eip712_with_array'
+        else pytest.mark.fail('unreachable code')
```
kclowes, line 267: sometimes you use `pytest.mark.fail` and sometimes `pytest.fail`. Is there a difference?

tests/core/test_structured_data_signing.py
```diff
@@ -34,124 +34,263 @@
...
+        ('eip712_with_array', 'Mail', ('Person',)),
+        ('eip712_with_array', 'Person', ()),
     )
 )
-def test_get_dependencies(primary_type, types, expected):
-    assert get_dependencies(primary_type, types) == expected
+def test_get_dependencies(
+    example, primary_type, eip712_example_types, eip712_example_with_array_types, expected
+):
+    _types = (
```
kclowes, line 128: At one point, we were moving away from having `if` statements in our tests for this sort of branching logic. I think it's a little clearer to just test `eip712_with_array` in a separate test, but I don't have a strong opinion on that if you like this better.

kclowes (approved):
Thank you for adding these! I just had a few nitpicky comments :) feel free to take or leave!

tests/core/test_structured_data_signing.py
fselmo: Agreed. I tried to make it so we can clearly see a path with `if` and basically an `elif` and otherwise purposely failing so that we don't get any false positives. I think this prevents the `if / else` from accidentally marking a test as `passed` which is arguably the main driver behind not having `if` statements in tests...

That said, I think separating it would be a lot better practice to be honest. I'll refactor this and it should resolve the comment on the `pytest.fail` as well. 👍

fselmo:
> Thank you for adding these! I just had a few nitpicky comments :) feel free to take or leave!

Good suggestion 👌. Updated. Mind taking a quick peek just for a last check since they were significant edits?

kclowes (approved):
LGTM! Thank you! 🚢

PR #154, opened 2022-04-06
improve bitarray version boundaries

ilanschnell:
As of version 2.4.0, bitarray is providing wheels for all types of platforms and Python versions on PyPI.  
Removing the upper version bound will make eth-account easier to install, as these wheels can be taken advantage of.  Bitarray is a very mature project with a very stable API, so there is no need for an upper version bound.


kclowes:
Nice! Thanks for the PR and for writing/maintaining bitarray! It seems like a better option may be to pin to `>=v2.4.0,<3`, as it looks like bitarray follows semantic versioning?

ilanschnell:
Sure, `>=v2.4.0,<3` will also be fine.  Yes, bitarray follows semantic versioning.

kclowes (approved):
I got around to testing it, and looks like we're all good here. Thanks @ilanschnell! I'll add a newsfragment and merge!

PR #155, opened 2022-04-06
Add python version testing to lint jobs, update mypy

kclowes:
## What was wrong?
`pip install -e .[dev]` on Python 3.9 and 3.10 wasn't working because mypy v0.770 isn't compatible.

## How was it fixed?

Updated `mypy`, and added lint testing for all supported versions of Python. This will need to be merged after #156. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
lgtm 👌

PR #156, opened 2022-04-07
Fix doc dependencies

kclowes:
## What was wrong?

Doc builds started failing.

## How was it fixed?

- Added a Jinja2 requirement of >=3.0,<3.1.0. Jinja3.1 breaks our docs with the following error: [ImportError: cannot import name 'contextfunction' from 'jinja2'](https://app.circleci.com/pipelines/github/ethereum/eth-account/197/workflows/41e90459-34c3-4c1e-a143-910b2d849963/jobs/4859)
  - See also: https://github.com/readthedocs/readthedocs.org/issues/9037
- Pinned towncrier to a certain version since they don't follow semver
- Tested higher sphinx versions and opened up the sphinx version requirement

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,8 @@ extras_require = {
         "pydocstyle>=5.0.0,<6",
     ],
     'doc': [
-        "Sphinx>=1.6.5,<2",
+        "Sphinx>=1.6.5,<5",
+        "jinja2>=3.0.0,<3.1.0",
         "sphinx_rtd_theme>=0.1.9,<1",
         "towncrier>=19.2.0, <20",
     ],
--- a/setup.py
+++ b/setup.py
@@ -20,9 +20,9 @@ extras_require = {
     ],
     'doc': [
         "Sphinx>=1.6.5,<5",
-        "jinja2>=3.0.0,<3.1.0",
+        "jinja2>=3.0.0,<3.1.0",  # jinja2<3.0 or >=3.1.0 cause doc build failures.
         "sphinx_rtd_theme>=0.1.9,<1",
-        "towncrier>=19.2.0, <20",
+        "towncrier==19.2.0",  # towncrier doesn't follow semver
     ],
     'dev': [
         "bumpversion>=0.5.3,<1",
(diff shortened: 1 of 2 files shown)


pacrob (approved):
lgtm

setup.py
```diff
@@ -19,9 +19,10 @@
         "pydocstyle>=5.0.0,<6",
     ],
     'doc': [
-        "Sphinx>=1.6.5,<2",
+        "Sphinx>=1.6.5,<5",
+        "jinja2>=3.0.0,<3.1.0",  # jinja2<3.0 or >=3.1.0 cause doc build failures.
         "sphinx_rtd_theme>=0.1.9,<1",
-        "towncrier>=19.2.0, <20",
+        "towncrier==18.5.0",  # towncrier doesn't follow semver
```
carver, line 25: Oh no, say it aint so! Did towncrier confirmed this in their docs somewhere?

Is there some interdependency that fails on more recent versions when towncrier is more recent? Just curious which API is changing that we are using.

I also got bitten by this issue in working on eth-portal. (I haven't run into a towncrier issue yet, but I also haven't generated any release notes) See: https://github.com/carver/eth-portal/commit/6ad07337dddc982ecf54de76c5c63874dccedd24

I suppose all these doc dependencies should get pinned here, too, then:
https://github.com/ethereum/ethereum-python-project-template/blob/b88fcf3c0b70ea7e4bd0fbc89b546a37877ce50e/setup.py#L20-L23
(whether the version you picked here, or the path I took)

(Plus, the commentary about why, and when the dependencies can change again.)

kclowes: > Did towncrier confirmed this in their docs somewhere?

No, but if you look at their [release history on pypi](https://pypi.org/project/towncrier/#history) you can see it used to be `year.month.something` and now I'm not sure what the pattern is.

The problem for us is that more recent versions than `18.5.0` were adding a double header on our releases. I suspect because it wasn't respecting the custom format that we had. 

> I suppose all these doc dependencies should get pinned here, too

good thought!

carver: > The problem for us is that more recent versions than `18.5.0` were adding a double header on our releases. I suspect because it wasn't respecting the custom format that we had.

Oh interesting, that's what it was!

carver: https://github.com/twisted/towncrier/blob/master/NEWS.rst#bugfixes

Looks like it was fixed in 21.9.0!

Issue #148, opened 2022-03-04
Bump bitarray dependency


lcsvcn:
Latest bitarray is 2.4.0. eth-account is still using < 1.3.0

It is good to keep lib update, for security and stability reasons.

```
The conflict is caused by:
    The user requested bitarray==2.4.0
    eth-account 0.6.1 depends on bitarray<1.3.0 and >=1.2.1

```

sudhamjayanthi:
+1

due to the old version, it's even causing me issues while trying to install eth_account.

here's what bitarray creator said - 

![image](https://user-images.githubusercontent.com/67156051/161749983-b24ccbff-c3dd-44d3-9e6f-be66bbe96e57.jpeg)

kclowes:
Closed in #163

PR #159, opened 2022-05-16
upgraded bitarray

dbfreem:
## What was wrong?

Issue #148 

## How was it fixed?

Hey @kclowes  I am leaving this as WIP at the moment but I upgraded the bitarray to the 2.4+ version and all the test run. I was going to look more at the code and see if I see anything breaking but I don't think there is. This will help a lot of the install issues on web3.py because wheels are available now for bitarray and no need to build it on install. Thoughts??

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         "hdaccount/wordlist/*.txt",
     ]},
     install_requires=[
-        "bitarray>=1.2.1,<1.3.0",
+        "bitarray>=2.4.0,<2.6",
         "eth-abi>=3.0.0,<4",
         "eth-keyfile>=0.6.0,<0.7.0",
         "eth-keys>=0.4.0,<0.5",


dbfreem:
Here is the test I ran locally.

1. Bumped eth-account to 0.6.2 and installed it in a virtual env
2. Since the requirement in web3.py is =>0.6.0,<0.7.0 no change was done on the web3.py side
3. I installed web3.py in the virtual env.
4. I verified that it used the eth-account 0.6.2 that was already installed in the virtual env and that there were no dependency conflict.
5. Ran py39-core test and everything passed.

Can you explain a little more about what you mean by this is a breaking change. At first glance I didn't see where this would break anything.

carver:
> It's a breaking change, but that's okay since we are in web3.py v6 beta.

If you don't feel like cutting a breaking release of eth-account, you could also just leave the lower limit at v1. (I think the upper limit should be v3, too)

Also, don't forget the newsfragment explaining the breaking change, in the PR. Speaking of which, something went [pretty sideways in the release notes](https://eth-account.readthedocs.io/en/stable/release_notes.html#eth-account-v0-6-1-2022-02-24) (double-release titles). I've seen it before, but can't remember how it got fixed. Maybe an update to towncrier?

carver:
Wow, lots of folks trying to make the same change at the same time and not noticing each other... See the earlier #154 

I'd recommend closing this in favor of #154 (after my suggested change above)

kclowes:
Thanks @dbfreem. I am going to close this one in favor of #154 as @carver suggested. This one (and all the others :) ) fell off the priority list because we were going to see if we could remove the bitarray dependency entirely, but this is a good option in the meantime and maybe this is a good enough fix for the foreseeable future. IIRC, the problem only shows up on windows machines. 

> Can you explain a little more about what you mean by this is a breaking change

Any time we update dependency versions, it's a breaking change. For example, if one of our dependencies requires `foo>=1.0.0,<2` and then we change our requirement from `foo>=1.0.0,<2` to `foo>=2,<3`, pip won't know how to resolve those two, and the dev environment will break for the user. 


> If you don't feel like cutting a breaking release of eth-account, you could also just leave the lower limit at v1. 

True. I think since we're in this weird in-between, maybe we make both changes. Have the breaking in the latest eth-account, and web3.py v6, and open it up to allow `>=1.0,<3` in the "stable" eth-account that's being used in web3.py v5.

dbfreem:
@kclowes I see what your are saying about the breaking change now. Thanks.

Issue #160, opened 2022-05-19
Support Python 3.11


oberstet:
## What was wrong?

Doesn't install on Python 3.11 from master branch

### Code that produced the error

```
pip install .
```

### Full error output 

```sh
(cpy311b1_1) (base) oberstet@intel-nuci7:~/scm/crossbario/eth-account$ pip install .
Processing /home/oberstet/scm/crossbario/eth-account
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting eth-keys<0.5,>=0.4.0
  Using cached eth_keys-0.4.0-py3-none-any.whl (21 kB)
Collecting bitarray<1.3.0,>=1.2.1
  Using cached bitarray-1.2.2.tar.gz (48 kB)
  Preparing metadata (setup.py) ... done
Requirement already satisfied: hexbytes<1,>=0.1.0 in /home/oberstet/cpy311b1_1/lib/python3.11/site-packages (from eth-account==0.6.1) (0.2.2)
Requirement already satisfied: eth-utils<3,>=2.0.0 in /home/oberstet/cpy311b1_1/lib/python3.11/site-packages (from eth-account==0.6.1) (2.0.0)
Collecting eth-keyfile<0.7.0,>=0.6.0
  Using cached eth_keyfile-0.6.0-py3-none-any.whl (6.6 kB)
Requirement already satisfied: eth-abi<4,>=3.0.0 in /home/oberstet/cpy311b1_1/lib/python3.11/site-packages (from eth-account==0.6.1) (3.0.0)
Collecting eth-rlp<1,>=0.3.0
  Using cached eth_rlp-0.3.0-py3-none-any.whl (5.0 kB)
Collecting rlp<4,>=1.0.0
  Using cached rlp-3.0.0-py2.py3-none-any.whl (20 kB)
Requirement already satisfied: eth-typing<4.0.0,>=3.0.0 in /home/oberstet/cpy311b1_1/lib/python3.11/site-packages (from eth-abi<4,>=3.0.0->eth-account==0.6.1) (3.1.0)
Requirement already satisfied: parsimonious<0.9.0,>=0.8.0 in /home/oberstet/cpy311b1_1/lib/python3.11/site-packages (from eth-abi<4,>=3.0.0->eth-account==0.6.1) (0.8.1)
[...]

PR #161, opened 2022-05-19
bump version limits for bitarray dependency

oberstet:
fixes https://github.com/ethereum/eth-account/issues/160

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         "hdaccount/wordlist/*.txt",
     ]},
     install_requires=[
-        "bitarray>=2.6.0,<3",
+        "bitarray>=2.5.1,<3",
         "eth-abi>=3.0.0,<4",
         "eth-keyfile>=0.6.0,<0.7.0",
         "eth-keys>=0.4.0,<0.5",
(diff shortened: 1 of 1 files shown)


carver:
Bumping this is a breaking change by our semver rules. If we left the lower limit and increase *only* the upper limit, then I expect it to be fine. Already installed copies should keep working fine, and newly installed copies will grab the latest available.

carver:
Ah, looks like this is already covered by #159 -- although I think you're right that the upper limit should just be v3

I'm going to close it as a duplicate, though

oberstet:
ok, sounds good! thanks!

> Bumping this is a breaking change by our semver rules. If we left the lower limit and increase only the upper limit

fwiw, in general, any upper limit is a pain. I have spent quite some hours because of a diamond shaped dependency graph with the same library being a dependency via two intermediary nodes, merging in a single app as tree root. if the dependency is required in different upper versions in the intermediary nodes in the dependency graph, pip will usually go into infinite looping to figure out a resolved tree. The solution then is to manually figure out a working version, pin that version in requirements of the app, so that pip will be forced to use that version. even figuring out that pip behavior took me many hours. IOW: pls try to avoid any upper limits;) my 2cts.

PR #163, opened 2022-05-20
Open up bitarray dependency to allow <3

kclowes:
## What was wrong?

Some users were have trouble installing web3.py because of the strict bitarray dependency requirement. In v2.4.0 bitarray added wheels for all platform types, which should fix the issue. This is the backwards compatible change, and the non-backwards compatible change (`>=2.4,<3`) was PRed against master.

Issue #154, #159

## How was it fixed?

See above.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
Makes sense to me 👍

Issue #101, opened 2020-04-22
Enable stricter mypy checks


carver:
## What was wrong?

Missing and/or wrong type definitions.

## How can it be fixed?

All of the strictness checks in mypy.ini should be set to `True`, like:
https://github.com/ethereum/ethereum-python-project-template/blob/582e9aa475bb56befb9a56188d2d33d176442b19/mypy.ini

There is a decent amount of work to do to add type annotations. There are a few wrong definitions too. Fix those up, and then the stricter checks will enforce it going forward.

PR #164, opened 2022-05-23
Switch on mypy disallow_incomplete_defs

Elnaril:
## What was wrong?

Issue #101: Enable stricter mypy checks

## How was it fixed?

Hello ! :)
I'm interested in this issue and would like to contribute if it's fine for you ?
As there is a bit of work and having the review process in mind, I would proceed step by step, by switching on 1 mypy flag at a time, if that sounds good to you ?
So here is the first one: disallow_incomplete_defs. Typing fixed soon™!

### To-Do

[//]: # (Work in progress)
- [x] Fix mypy

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -3,6 +3,13 @@ from collections.abc import (
 )
 import json
 import os
+from typing import (
+    Optional,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+)
 import warnings
 
 from cytoolz import (
@@ -18,6 +25,11 @@ from eth_keys import (
 )
 from eth_keys.exceptions import (
     ValidationError,
+)
+from eth_typing import (
+    ChecksumAddress,
+    Hash32,
+    HexStr,
 )
 from eth_utils.curried import (
     combomethod,
@@ -63,6 +75,8 @@ from eth_account.messages import (
 from eth_account.signers.local import (
     LocalAccount,
 )
+
+VRS = TypeVar('VRS', bytes, HexStr, int)
 
 
 class Account(object):
@@ -254,7 +268,7 @@ class Account(object):
     def from_mnemonic(self,
                       mnemonic: str,
                       passphrase: str = "",
-                      account_path: str = ETHEREUM_DEFAULT_PATH):
+                      account_path: str = ETHEREUM_DEFAULT_PATH) -> LocalAccount:
         """
         Generate an account from a mnemonic.
 
@@ -324,7 +338,7 @@ class Account(object):
                              passphrase: str = "",
                              num_words: int = 12,
                              language: str = "english",
-                             account_path: str = ETHEREUM_DEFAULT_PATH):
+                             account_path: str = ETHEREUM_DEFAULT_PATH) -> Tuple[LocalAccount, str]:
         r"""
         Create a new private key and related mnemonic.
 
@@ -367,7 +381,10 @@ class Account(object):
         return self.from_mnemonic(mnemonic, passphrase, account_path), mnemonic
 
     @combomethod
-    def recover_message(self, signable_message: SignableMessage, vrs=None, signature=None):
+    def recover_message(self,
+                        signable_message: SignableMessage,
+                        vrs: Optional[Tuple[VRS, VRS, VRS]] = None,
+                        signature: bytes = None) -> ChecksumAddress:
         r"""
         Get the address of the account that signed the given message.
         You must specify exactly one of: vrs or signature
@@ -431,7 +448,7 @@ class Account(object):
             '0x5ce9454909639D2D17A3F753ce7d93fa0b9aB12E'
         """
         message_hash = _hash_eip191_message(signable_message)
-        return self._recover_hash(message_hash, vrs, signature)
+        return cast(ChecksumAddress, self._recover_hash(message_hash, vrs, signature))
 
     @combomethod
     def recoverHash(self, message_hash, vrs=None, signature=None):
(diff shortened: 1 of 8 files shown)


Elnaril:
Hello!

### Tox issue
While working on this PR, I noticed that tox was not running the lint section. For example, we can see the actual executions are missing after the "run-test-pre":

image

I believe this is due to 
https://github.com/ethereum/eth-account/blob/43f8cfb4637e64318846cd85f36dfcb07aefb29e/tox.ini#L41

To fix that I would replace:
https://github.com/ethereum/eth-account/blob/43f8cfb4637e64318846cd85f36dfcb07aefb29e/tox.ini#L41-L43

with:
image

Which means updating tox version to at least 3.15, according to my understanding of the [doc](https://tox.wiki/en/latest/config.html#generative-section-names)

And now mypy is launched by tox:
image

I'm not familiar with circleci and its config.yml, but I guess it works differently since it executes the lint section automatically.
I also noticed circleci uses the version 3.25 of tox, so we could upgrade tox to the same version in setup.py.

### Mypy issue
This leads to the following mypy issue:
image

due to:
https://github.com/ethereum/eth-account/blob/43f8cfb4637e64318846cd85f36dfcb07aefb29e/eth_account/hdaccount/deterministic.py#L78-L79

According to the referenced [typeshed issue](https://github.com/python/typeshed/issues/2686), the problem is fixed and indeed, removing the `#type: ignore` comment fixes mypy:

image

image

Though, I'm not sure why the issue is not raised by CircleCI ... Maybe because of using a cache ?

### Conclusion
If that sounds good to you, I'm happy to include these changes in this PR or create a specific PR, which ever you prefer ?

kclowes:
> As there is a bit of work and having the review process in mind, I would proceed step by step, by switching on 1 mypy flag at a time, if that sounds good to you ?

That would be great, thank you!

> While working on this PR, I noticed that tox was not running the lint section

Oops, thank you. We would very much appreciate the fix you posted. 

> I'm happy to include these changes in this PR or create a specific PR, which ever you prefer ?

Smaller PRs are easier to review, so the smaller the better. Would love to have a separate PR for getting the lint going on CI again, and another for turning on the flag. Thanks!

Elnaril:
Sure, will do! :)

Elnaril:
Would it be possible to run the CI on this PR as well ?

eth_account/account.py
```diff
@@ -539,7 +558,9 @@ def set_key_backend(self, backend):
         self._keys = KeyAPI(backend)
 
     @combomethod
-    def sign_message(self, signable_message: SignableMessage, private_key):
+    def sign_message(self,
+                     signable_message: SignableMessage,
+                     private_key: Union[bytes, HexStr, int, keys.PrivateKey]) -> Any:
```
Elnaril, line 563: Any i/o SignedMessage because of decorator

eth_account/account.py
```diff
@@ -614,7 +635,9 @@ def signHash(self, message_hash, private_key):
         return self._sign_hash(message_hash, private_key)
 
     @combomethod
-    def _sign_hash(self, message_hash, private_key):
+    def _sign_hash(self,
+                   message_hash: Hash32,
+                   private_key: Union[bytes, HexStr, int, keys.PrivateKey]) -> Any:
```
Elnaril, line 640: Any i/o SignedMessage because of decorator

Elnaril:
I've re-based and cleaned the commits, added a release note, tox runs fine locally, so I think this PR is ready for review! :)

eth_account/account.py
```diff
@@ -367,7 +380,10 @@ def create_with_mnemonic(self,
         return self.from_mnemonic(mnemonic, passphrase, account_path), mnemonic
 
     @combomethod
-    def recover_message(self, signable_message: SignableMessage, vrs=None, signature=None):
+    def recover_message(self,
+                        signable_message: SignableMessage,
+                        vrs: Optional[Tuple[VRS, VRS, VRS]] = None,
+                        signature: bytes = None) -> Any:  # i/o ChecksumAddress due to decorator.
```
kclowes, line 386: Instead of leaving this as `Any`, we can change it to be `ChecksumAddress`, and then change line 474 to be a cast:

`return cast(ChecksumAddress, self._recover_hash(message_hash, vrs, signature)`

eth_account/account.py
```diff
@@ -458,7 +474,10 @@ def recoverHash(self, message_hash, vrs=None, signature=None):
         return self._recover_hash(message_hash, vrs, signature)
 
     @combomethod
-    def _recover_hash(self, message_hash, vrs=None, signature=None):
+    def _recover_hash(self,
+                      message_hash: Hash32,
+                      vrs: Optional[Tuple[VRS, VRS, VRS]] = None,
+                      signature: bytes = None) -> Any:  # i/o ChecksumAddress due to decorator.
```
kclowes, line 480: Same here re. changing `Any` -> `ChecksumAddress`, and casting the return value.

eth_account/account.py
kclowes: same here:
```suggestion
                     private_key: Union[bytes, HexStr, int, keys.PrivateKey]) -> SignedMessage:
```
 then `return cast(SignedMessage, self._sign_hash(message_hash, private_key)` on line 607

eth_account/account.py
kclowes: same here:
```suggestion
                   private_key: Union[bytes, HexStr, int, keys.PrivateKey]) -> SignedMessage:
```

kclowes (commented):
Thank you! I had a few comments around getting rid of the `Any` return type for combomethods and `cast`ing the return values instead. I think it's a little bit better, but open to other opinions if you have them :) looks great otherwise!

Elnaril:
Thanks a lot for your feedback! I've applied your suggestions and they work like a charm! :)
I've squashed my commits and tox is happy locally, so, ready for last round of review!

kclowes (approved):
This looks great! Thanks for taking it on! I'll ~add a newsfragment and~ merge

Edit: just saw there is a newsfragement already. Thanks!

Elnaril:
Nice! :)
I'll carry on with the next flag when I can free up some time! Thanks!

PR #167, opened 2022-05-26
Added sudo to pip command README.md

LalleSX:
not running sudo will result in errors

## What was wrong?
running without sudo had errors, [Kubuntu LTS 20.4] 
Issue #

## How was it fixed?
sudo was added
Summary of approach.
sudo was added to command
### To-Do

- [ ] Clean up commit history

- [ ] Add entry to the [release notes](https://github.com/ethereum/eth-account/blob/master/newsfragments/README.md)

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Read more in the [documentation on ReadTheDocs](https://eth-account.readthedocs.
 ## Quickstart
 
 ```sh
-pip install eth-account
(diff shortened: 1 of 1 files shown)


kclowes:
Thanks for the PR, but `pip install` should work without `sudo`. I suspect there is something specific to your environment that requires a `sudo`.

PR #170, opened 2022-06-05
Switch on mypy flag disallow_any_generics

Elnaril:
## What was wrong?

Issue https://github.com/ethereum/eth-account/issues/101: Enable stricter mypy checks

## How was it fixed?

I switched on the mypy flag "disallow_any_generics", and fixed resulting mypy errors.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_account/_utils/transaction_utils.py
+++ b/eth_account/_utils/transaction_utils.py
@@ -1,9 +1,16 @@
 from typing import (
     Any,
     Dict,
+    Mapping,
     Sequence,
+    Tuple,
+    Union,
+    cast,
 )
 
+from eth_typing import (
+    ChecksumAddress,
+)
 from toolz import (
     assoc,
     dissoc,
@@ -13,6 +20,9 @@ from eth_account._utils.validation import (
     is_rlp_structured_access_list,
     is_rpc_structured_access_list,
 )
+
+RLPStructure = Tuple[Union[bytes, bytearray, ChecksumAddress], Tuple[Union[int, str]]]
+RPCStructure = Tuple[Dict[str, Union[bytes, bytearray, ChecksumAddress, int, str]]]
 
 
 def set_transaction_type_if_needed(transaction_dict: Dict[str, Any]) -> Dict[str, Any]:
@@ -39,7 +49,7 @@ def transaction_rpc_to_rlp_structure(dictionary: Dict[str, Any]) -> Dict[str, An
     return dictionary
 
 
-def _access_list_rpc_to_rlp_structure(access_list: Sequence) -> Sequence:
+def _access_list_rpc_to_rlp_structure(access_list: Sequence[Mapping[str, Any]]) -> RLPStructure:
     if not is_rpc_structured_access_list(access_list):
         raise ValueError("provided object not formatted as JSON-RPC-structured access list")
     rlp_structured_access_list = []
@@ -51,7 +61,7 @@ def _access_list_rpc_to_rlp_structure(access_list: Sequence) -> Sequence:
                 tuple(_ for _ in d['storageKeys'])  # tuple of storage key values
             )
         )
-    return tuple(rlp_structured_access_list)
+    return cast(RLPStructure, tuple(rlp_structured_access_list))
 
 
 # rlp to JSON-RPC transaction structure
@@ -67,7 +77,7 @@ def transaction_rlp_to_rpc_structure(dictionary: Dict[str, Any]) -> Dict[str, An
     return dictionary
 
 
-def _access_list_rlp_to_rpc_structure(access_list: Sequence) -> Sequence:
+def _access_list_rlp_to_rpc_structure(access_list: Sequence[Sequence[Any]]) -> RPCStructure:
     if not is_rlp_structured_access_list(access_list):
         raise ValueError("provided object not formatted as rlp-structured access list")
     rpc_structured_access_list = []
@@ -79,4 +89,4 @@ def _access_list_rlp_to_rpc_structure(access_list: Sequence) -> Sequence:
                 'storageKeys': t[1]
             }
         )
-    return tuple(rpc_structured_access_list)
+    return cast(RPCStructure, tuple(rpc_structured_access_list))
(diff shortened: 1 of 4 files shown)


Elnaril:
Hello! :)
Can someone please trigger the CI ?

kclowes:
Done! I'll need to check to see if CI isn't running on purpose 🤔 thanks for the PR!

Elnaril:
Hi!
Can someone please trigger the CI ?

kclowes:
Done! I just did some mucking around with the CircleCI settings, so hopefully it'll just build automatically now 🤞

Elnaril:
Hi @kclowes ! I've just cleaned my commits so this PR is ready for review! :)
Though the CI was not triggered

eth_account/_utils/transaction_utils.py
```diff
@@ -51,7 +61,7 @@ def _access_list_rpc_to_rlp_structure(access_list: Sequence) -> Sequence:
                 tuple(_ for _ in d['storageKeys'])  # tuple of storage key values
```
kclowes, line 61: I can't leave a comment on line 50, but we may need to `cast(ChecksumAddress, d['address'])` there to get the other types working. If you can figure out how to do it without the cast, that would be even better!

eth_account/_utils/transaction_utils.py
```diff
@@ -51,7 +61,7 @@ def _access_list_rpc_to_rlp_structure(access_list: Sequence) -> Sequence:
...
             )
         )
-    return tuple(rlp_structured_access_list)
+    return cast(RLPStructure, tuple(rlp_structured_access_list))
```
kclowes, line 64: Then I think we can remove this `cast`

eth_account/_utils/transaction_utils.py
```diff
@@ -79,4 +89,4 @@ def _access_list_rlp_to_rpc_structure(access_list: Sequence) -> Sequence:
                 'storageKeys': t[1]
             }
         )
-    return tuple(rpc_structured_access_list)
+    return cast(RPCStructure, tuple(rpc_structured_access_list))
```
kclowes, line 92: And I think we can also remove this `cast`

eth_account/_utils/transaction_utils.py
```diff
@@ -14,6 +21,9 @@
     is_rpc_structured_access_list,
 )
 
+RLPStructure = Tuple[Union[bytes, bytearray, ChecksumAddress], Tuple[Union[int, str]]]
+RPCStructure = Tuple[Dict[str, Union[bytes, bytearray, ChecksumAddress, int, str]]]
```
kclowes, line 25: And this can be:
```suggestion
RPCStructure = Tuple[Dict[str, Tuple[Hash32, ...]], ...]
```

eth_account/_utils/transaction_utils.py
```diff
@@ -14,6 +21,9 @@
     is_rpc_structured_access_list,
 )
 
+RLPStructure = Tuple[Union[bytes, bytearray, ChecksumAddress], Tuple[Union[int, str]]]
```
kclowes, line 24: I think this can be:
```suggestion
RLPStructure = Tuple[Tuple[str, Tuple[Hash32, ...]], ...]
```

kclowes:
I was just reading through some docs [here](https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories), and the second note in that section says: 
> If a user submits a pull request to your repository from a fork, but no pipeline is triggered, then the user most likely is following a project fork on their personal account rather than the project itself of CircleCi, causing the jobs to trigger under the user’s personal account and not the organization account. To resolve this issue, have the user unfollow their fork of the project on CircleCI and instead follow the source project. This will trigger their jobs to run under the organization when they submit pull requests.

So maybe try following the ethereum/eth-account project on CircleCI to get the tests to run?

kclowes (commented):
Thank you @Elnaril! I left some comments around the `RLPStructure` and RPCStructure definitions that I think will allow us remove some of the `cast`ing. Let me know if you have questions!

Elnaril:
Hi @kclowes ! Thank you for your hints and feedbacks, and sorry for the late answer!
I have a few questions:


1. According to your suggestion to define: 
```python
RLPStructure = Tuple[Tuple[str, Tuple[Hash32, ...]], ...]
```
I understand the expected address is always a `str` (or a `ChecksumAddress` which is a `str`).
But I also understand the function `is_rlp_structured_access_list()`, that checks the `access_list` validity, allows also `bytes` and `bytearray`.
See: [`is_rlp_structured_access_list()`](https://github.com/ethereum/eth-account/blob/7f5eefccaf4617e645895106bf396e728f81f64e/eth_account/_utils/validation.py#L87-L88), and: [`is_address()`](https://github.com/ethereum/eth-utils/blob/db5d80a7efc4b00cf759721c2a9ee6ea2edb5392/eth_utils/address.py#L44-L45)
So, my questions are:
 - are we sure we always get addresses as str ?
 - if yes, wouldn't it make sense to restrict the validity check and allow only `str`/`ChecksumAddress` ?

Same considerations for the keys for which the validity check seems to allow `int`s and not only `bytes`/`Hash32` ?


2. Based on you suggestions, it seems we can solve the `cast` issue if we use a `TypedDict` as follow:

```python
RLPStructure = Tuple[Tuple[str, Tuple[Hash32, ...]], ...]
RPCDict = TypedDict('RPCDict', {'address': str, 'storageKeys': Tuple[Hash32, ...]})
RPCStructure = Tuple[RPCDict, ...]
```
I've committed this change and the CI is fine except for Python 3.7 as `TypedDict` was added to the standard library only in 3.8.

So, my question is: would it be acceptable to use `typing_extensions` like in [web3.py](https://github.com/ethereum/web3.py/blob/master/web3/_utils/compat/__init__.py) ? That would mean to add a dependency in `setup.py` since only `mypy` uses it at the moment.

pacrob:
closed by #285

PR #178, opened 2022-06-23
fix redos-able regex

pacrob:
## What was wrong?

We were notified that a regex we are using was DoS-able
https://en.wikipedia.org/wiki/ReDoS

## How was it fixed?

Fixed it so that the demonstrated vulnerability no longer causes problems.

### To-Do

#### Cute Animal Picture
image


kclowes (approved):
Looks good to me! Thanks for taking this on!

PR #202, opened 2022-09-27
Drop `eth-abi` dependency upper bound

fselmo:
## What was wrong?

- ``eth-abi 4.0.0-beta.1`` was released with a crucial fix to zero-padding. In an effort to loosen restrictions, and to be able allows projects using `eth-account` to fine tune their own version of `eth-abi`, drop the upper bound of the `eth-abi` dependency.

## How was it fixed?

- Remove upper bound for `eth-abi`; use `>=3.0.1`

### To-Do

- [x] Clean up commit history


kclowes (approved):
🚀 !

PR #216, opened 2023-01-22
replace first bunch of cytoolz helper usage

### LINKED ISSUE #215: Implicit dependency on cytoolz (breaks PyPy)
The library [depends](https://toolz.readthedocs.io/en/latest/api.html?highlight=dissoc#toolz.dicttoolz.dissoc) on `toolz.dicttoolz.dissoc`

https://github.com/ethereum/eth-account/blob/c5552a56b349e4c205089f50ddf54e942d4d9bc9/eth_account/account.py#L752

even though this requirement is *not* stated in the package setup

https://github.com/ethereum/eth-account/blob/c5552a56b349e4c205089f50ddf54e942d4d9bc9/setup.py#L67

This could be replaced with

```python
from copy import deepcopy
sanitized_transaction = deepcopy(transaction_dict)
sanitized_transaction.pop('from', None)
```

without introducing a dependency on cytoolz and allowing this library to be used on PyPy.

oberstet:
fixes https://github.com/ethereum/eth-account/issues/215

### DIFF
--- a/eth_account/_utils/legacy_transactions.py
+++ b/eth_account/_utils/legacy_transactions.py
@@ -1,15 +1,8 @@
 import itertools
-from copy import deepcopy
 from typing import (
     Dict,
 )
 
-from toolz import (
-    curry,
-    merge,
-    partial,
-    pipe,
-)
 from eth_rlp import (
     HashableRLP,
 )
@@ -21,6 +14,15 @@ from rlp.sedes import (
     Binary,
     big_endian_int,
     binary,
+)
+from toolz import (
+    curry,
+    merge,
+    partial,
+    pipe,
+)
+from toolz.dicttoolz import (
+    dissoc,
 )
 
 from .transaction_utils import (
@@ -58,11 +60,7 @@ def serializable_unsigned_transaction_from_dict(transaction_dict):
 
 def encode_transaction(unsigned_transaction, vrs):
     (v, r, s) = vrs
-    chain_naive_transaction = deepcopy(unsigned_transaction.as_dict())
-    for key in ["v", "r", "s"]:
-        if key in chain_naive_transaction:
-            del chain_naive_transaction[key]
-
+    chain_naive_transaction = dissoc(unsigned_transaction.as_dict(), "v", "r", "s")
     if isinstance(unsigned_transaction, TypedTransaction):
         # Typed transaction have their own encoding format,
         # so we must delegate the encoding.
--- a/eth_account/_utils/legacy_transactions.py
+++ b/eth_account/_utils/legacy_transactions.py
@@ -1,11 +1,11 @@
 import itertools
+from copy import deepcopy
 from typing import (
     Dict,
 )
 
-from cytoolz import (
+from toolz import (
     curry,
-    dissoc,
     merge,
     partial,
     pipe,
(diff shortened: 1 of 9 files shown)


oberstet:
the CI is currently failing with the following - which seems unrelated to this PR:

```
## Confirming "jammy" is supported...

+ curl -sLf -o /dev/null 'https://deb.nodesource.com/node_12.x/dists/jammy/Release'

## Your distribution, identified as "jammy", is not currently supported, please contact NodeSource at https://github.com/nodesource/distributions/issues if you think this is incorrect or would like your distribution to be considered for support
```

grafik

eth_account/_utils/legacy_transactions.py
```diff
@@ -22,6 +15,15 @@
     big_endian_int,
     binary,
 )
+from toolz import (
```
kclowes, line 18: For example, this should be:
```py
from eth_utils import (
    curry,
    ...
)
```

kclowes (commented):
Thanks for this PR! You can actually import the methods from `eth-utils` which will check for the method in `cytoolz` and fall back to `toolz`. 

You're right that the CI failures are unrelated, I believe we need to bump our Node version for the integration tests.

oberstet:
> You can actually import the methods from `eth-utils`

Yeah, I could, but why more complicated? Using toolz works fine for me on both CPython and PyPy.

I have spent countless hours with the Ethereum foundation Python packages fighting dependency issues.

All the Ethereum foundation Python packages unfortunately specify *upper bounds* on package dependencies.

Since Pip/Setuptools is unable to cope with the resulting confluent dependencies in the dependency graph of apps, this results in lots of pain and hours wasted in downstream projects.

As soon as your app (or any direct or indirect dependency) depends on 2 deps which both depend on 1 eth package - but insist on different versions of that shared dep (confluence in dependency graph) - you are up for lots of fun;) 

And I am talking about apps which use a properly isolated, app-specific venv! For distros shipping Python packages as system-wide distro packages, making this work for users and apps will become simply impossible.

IMO, the Ethereum Python packages, their dependencies policy that is, is severely broken, and I have to use forked versions of the libraries anyways, with upper bounds removed. End of rant. Sorry.

oberstet:
> Yeah, I could, but why more complicated?

just noticed, this library depends on `eth-utils` anyways already ... _with_ upper bound: means, using it shouldn't create _additional_ problems;)

https://github.com/ethereum/eth-account/blob/c5552a56b349e4c205089f50ddf54e942d4d9bc9/setup.py#L73

are you interested in merging this PR? then I would invest more work into the PR, continue use my fork only for killing the upper bounds ..

kclowes:
> Yeah, I could, but why more complicated? Using toolz works fine for me on both CPython and PyPy.

Cytoolz is more performant so we like to give people the option if possible. And like you said `eth-utils`  is already a dependency, so it shouldn't be more complicated.

> All the Ethereum foundation Python packages unfortunately specify upper bounds on package dependencies

We have been exploring removing the upper bounds on our dependencies across our libraries. Web3.py v6b8 has that change in already if you want to test it out. Pinning upper bounds and not pinning upper bounds both have pros and cons. But your feedback has been heard. 

> are you interested in merging this PR?

Yes. One of us can pick it up in a bit if you don't want to put any more time in but we'd appreciate the contribution. We're focused on getting web3.py v6 stable out now so it will probably take us some time to get to this.

oberstet:
> Cytoolz is more performant

On which Python implementation? Without measuring, I would expect it to be faster on CPython and slower on PyPy. The tracing JIT compiler of PyPy spills out native machine code. PyPy is usually faster with pure Python JIT'ed than a C compiler, unless using vectorized machine code (SIMD). In which case, Numba (also a JIT compiler, but non-tracing .. I think) will do the shootout with the C compilers.

Anyways: _if_ `eth-utils` does the right job, that is _not_ using cytoolz if not available or when running on PyPy, and falls back to toolz, cool!!! Doesn't harm then.

> Pinning upper bounds and not pinning upper bounds both have pros and cons. But your feedback has been heard.

Thanks a lot! Yeah, there are different opinions. FWIW, my opinion (which is what I try to follow wherever I can):

* don't use distro packaged Python packages
* use 1 dedicated venv for every application
* specify all deps of that app version / even better: hash pinned - that is stronger than upper bounds)
* specify lower bounds and no upper bounds in all libraries (Python package deps of the app)

> but we'd appreciate the contribution

ok, I'll get back to this hopefully tomorrow and do the changes as indicated by you.

if you have a fix for the Node version thing, I'd be glad to incorporate! as it is more fun to have proper "fully green" PR in the end;)

PR #222, opened 2023-03-13
remove notices of Draft status for eips 712 and 191, both now in Final

pacrob:
## What was wrong?

EIPs 712 and 191 have moved to Final status.

## How was it fixed?

Removed notes saying they were in Draft status.

### To-Do

#### Cute Animal Picture

image


kclowes (approved):
:shipit:

PR #223, opened 2023-03-15
fix node breaking ci

pacrob:
## What was wrong?

`npm install -g .` seems to have changed function. It creates a symlink back to the local folder (`tests/integration/ethers-cli` in this case) rather than installing all the dependencies locally.

## How was it fixed?

Added a local `npm install` that actually installs dependencies. Also pulled a new npm install script in.

### To-Do

#### Cute Animal Picture

image

### DIFF
--- a/tests/integration/ethers-cli/setup_node_v18.sh
+++ b/tests/integration/ethers-cli/setup_node_v18.sh
@@ -3,14 +3,14 @@
 # Discussion, issues and change requests at:
 #   https://github.com/nodesource/distributions
 #
-# Script to install the NodeSource Node.js 12.x repo onto a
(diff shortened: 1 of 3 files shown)


tests/integration/ethers-cli/setup_node_v18.sh
```diff
@@ -297,28 +330,35 @@ if [ -f "/etc/apt/sources.list.d/chris-lea-node_js-$DISTRO.list" ]; then
 fi
 
 print_status 'Adding the NodeSource signing key to your keyring...'
+keyring='/usr/share/keyrings'
```
kclowes, line 333: Nice, thanks for updating this!

kclowes (approved):
LGTM!

PR #224, opened 2023-03-21
bump docs versions and update .readthedocs.yml

pacrob:
## What was wrong?

Docs dependencies were scoped low

## How was it fixed?

Raised lower bar, removed upper pin

### To-Do

#### Cute Animal Picture

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,9 +21,8 @@ extras_require = {
         "black>=22,<23",
     ],
     "doc": [
-        "Sphinx>=1.6.5,<5",
-        "jinja2>=3.0.0,<3.1.0",  # jinja2<3.0 or >=3.1.0 cause doc build failures.
-        "sphinx_rtd_theme>=0.1.9,<1",
+        "sphinx>=5.0.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -21,9 +21,9 @@
         "black>=22,<23",
     ],
     "doc": [
-        "Sphinx>=1.6.5,<5",
-        "jinja2>=3.0.0,<3.1.0",  # jinja2<3.0 or >=3.1.0 cause doc build failures.
-        "sphinx_rtd_theme>=0.1.9,<1",
+        "Sphinx>=4.2.0",
+        "jinja2>=3.0.0",  # jinja2<3.0 or >=3.1.0 cause doc build failures.
```
kclowes, line 25: we may be able to remove this line in general?

pacrob: If we don't use Jinja at all, then for sure. I'll try it. This is where I went down the RTD webhook refresh rabbithole.

kclowes (approved):
🧹 Thanks! :shipit:

PR #228, opened 2023-04-26
Upgrade template

pacrob:
## What was wrong?

The project template has been updated, this pulls in most of the new changes.

### To-Do

- [ ] Clean up commit history

- [ ] Add entry to the [release notes](https://github.com/ethereum/eth-account/blob/master/newsfragments/README.md)

#### Cute Animal Picture

image


eth_account/signers/local.py
```diff
@@ -86,6 +86,7 @@ def signTransaction(self, transaction_dict):
         warnings.warn(
             "signTransaction is deprecated in favor of sign_transaction",
             category=DeprecationWarning,
+            stacklevel=2,
```
kclowes, line 89: I'm curious about this change - is it a `flake8-bugbear` catch?

kclowes (approved):
Looks good to me! Just had a few random comments.

eth_account/signers/local.py
pacrob: It is `flake8-bugbear`. The full error was:

```
 B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
```

It sounded reasonable to me so just added the `stacklevel` arg.

README.md
```diff
@@ -37,52 +37,11 @@ git clone git@github.com:ethereum/eth-account.git
 cd eth-account
 virtualenv -p python3 venv
 . venv/bin/activate
-pip install -e ".[dev]"
-```
-
-To run the integration test cases, you need to install node and the custom cli tool as follows:
```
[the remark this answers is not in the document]
pacrob: Yup, that's my oversight. Added back.

setup.py
```diff
@@ -57,40 +59,39 @@
...
-        "rlp>=1.0.0,<4",
+        "eth-keyfile>=0.6.0",
+        "eth-keys>=0.4.0",
+        "eth-rlp>=0.3.0",
+        "eth-utils>=2.0.0",
+        "hexbytes>=0.1.0",
+        "rlp>=1.0.0",
     ],
-    python_requires=">=3.6, <4",
+    python_requires=">=3.7, <4",
```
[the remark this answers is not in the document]
pacrob: added a breaking newsfragment

eth_account/signers/local.py
kclowes: Yep, makes sense to me! Thanks!

PR #221, opened 2023-03-08
add TypedDataConvert to message

marktohark:
## What was wrong?

I get sign json from blur.io
when sign json convert to py dict, I need to fix value type manually.

## How was it fixed?
Add a helper to message.py

```python
import json
from eth_account.messages import encode_structured_data, TypedDataConvert

s1 = '''{"type":"signTypedData_v4","data":{"domain":{"name":"Blur Exchange","version":"1.0","chainId":1,"verifyingContract":"0x000000000000ad05ccc4f10045630fb830b95127"},"types":{"Order":[{"name":"trader","type":"address"},{"name":"side","type":"uint8"},{"name":"matchingPolicy","type":"address"},{"name":"collection","type":"address"},{"name":"tokenId","type":"uint256"},{"name":"amount","type":"uint256"},{"name":"paymentToken","type":"address"},{"name":"price","type":"uint256"},{"name":"listingTime","type":"uint256"},{"name":"expirationTime","type":"uint256"},{"name":"fees","type":"Fee[]"},{"name":"salt","type":"uint256"},{"name":"extraParams","type":"bytes"},{"name":"nonce","type":"uint256"}],"Fee":[{"name":"rate","type":"uint16"},{"name":"recipient","type":"address"}]},"value":{"trader":"0xcb672216a6290ea85cdf75d3b8cbd18c99cac50e","side":0,"matchingPolicy":"0x0000000000b92d5d043faf7cecf7e2ee6aaed232","collection":"0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d","tokenId":"0","amount":"1","paymentToken":"0x0000000000a39bb272e79075ade125fd351887ac","price":"10000000000000000","listingTime":1678256751,"expirationTime":1709792748,"fees":[],"salt":"0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d61099c9f51e5d0332a4a44cf2aa174b","extraParams":"0x01","nonce":"0"}}}'''

s1 = json.loads(s1)

signInfo = TypedDataConvert(s1['data']['domain'], s1['data']['types'], s1['data']['value'])
encodeData = encode_structured_data(signInfo)
res = web3.auto.w3.eth.account.sign_message(encodeData, private_key=pk)
print(res.signature.hex())
```

PR #235, opened 2023-08-03
Add encode typed data

pacrob:
### What was wrong?

EIP712 message signing could be more user-friendly. eth-account's `encode_structured_data` has a different interface and encodes edge cases differently than users of Metamask and ethers.js have come to expect.

Closes #221

### How was it fixed?

Using Metamask's `eth-sig-util` and ethers.js as a model, I built out new mechanics for encoding EIP712 typed data.

Notable differences:
 - No attention paid to array dimensions
 - `None` values are allowed for custom, `string`, and `bytes` types.

Allowing `encode_typed_data` to be called with a single dict in the manner of `encode_structured_data` will be added in a separate PR.

### Todo:
- [x] rename `encode_structured_data` to indicate legacy and add deprecation notice
- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/_utils/encode_typed_data/encoding_and_hashing.py
+++ b/eth_account/_utils/encode_typed_data/encoding_and_hashing.py
@@ -3,6 +3,7 @@ from typing import (
     Dict,
     List,
     Tuple,
+    Union,
 )
 
 from eth_abi import (
@@ -12,46 +13,26 @@ from eth_utils import (
     is_hexstr,
     keccak,
     to_bytes,
+    to_int,
+)
+
+from eth_account._utils.encode_typed_data.helpers import (
+    EIP712_SOLIDITY_TYPES,
+    coerce_bool,
+    is_array_type,
+    parse_core_array_type,
+    parse_parent_array_type,
 )
 
 
-def _get_EIP712_solidity_types():
-    types = ["bool", "address", "string", "bytes"]
-    ints = [f"int{(x + 1) * 8}" for x in range(32)]
-    uints = [f"uint{(x + 1) * 8}" for x in range(32)]
-    bytes_ = [f"bytes{x + 1}" for x in range(32)]
-    return types + ints + uints + bytes_
-
-
-EIP712_SOLIDITY_TYPES = _get_EIP712_solidity_types()
-
-
-def _is_array_type(type_: str) -> bool:
-    return type_.endswith("]")
-
-
-# strip all brackets: Person[][] -> Person
-def _parse_core_array_type(type_: str) -> str:
-    if _is_array_type(type_):
-        type_ = type_[: type_.index("[")]
-    return type_
-
-
-# strip outermost brackets: Person[][] -> Person[]
-def _parse_inner_array_type(type_: str) -> str:
-    if _is_array_type(type_):
-        type_ = type_[: type_.rindex("[")]
-    return type_
-
-
-def _get_primary_type(types: Dict[str, List[Dict[str, str]]]) -> str:
+def get_primary_type(types: Dict[str, List[Dict[str, str]]]) -> str:
     custom_types = set(types.keys())
     custom_types_that_are_deps = set()
 
     for type in custom_types:
         type_fields = types[type]
         for field in type_fields:
-            parsed_type = _parse_core_array_type(field["type"])
+            parsed_type = parse_core_array_type(field["type"])
             if parsed_type in custom_types and parsed_type != type:
                 custom_types_that_are_deps.add(parsed_type)
 
@@ -67,7 +48,7 @@ def encode_field(
     name: str,
     type_: str,
     value: Any,
-) -> Tuple[str, bytes]:
+) -> Tuple[str, Union[int, bytes]]:
     if type_ in types.keys():
         # type is a custom type
         if value is None:
@@ -91,6 +72,15 @@ def encode_field(
     if value is None:
         raise ValueError(f"missing value for field {name} of type {type_}")
 
+    if is_array_type(type_):
+        type_ = parse_parent_array_type(type_)
+        type_value_pairs = [encode_field(types, name, type_, item) for item in value]
+        data_types, data_hashes = zip(*type_value_pairs)
+        return ("bytes32", keccak(encode(data_types, data_hashes)))
+
+    if type_ == "bool":
+        return (type_, coerce_bool(value))
+
     if type_ == "bytes":
         if is_hexstr(value):
             value = to_bytes(hexstr=value)
@@ -105,11 +95,12 @@ def encode_field(
             value = to_bytes(text=value)
         return ("bytes32", keccak(value))
 
-    if _is_array_type(type_):
-        type_ = _parse_inner_array_type(type_)
-        type_value_pairs = [encode_field(types, name, type_, item) for item in value]
-        data_types, data_hashes = zip(*type_value_pairs)
-        return ("bytes32", keccak(encode(data_types, data_hashes)))
+    # allow string values for int and uint types
+    if type(value) == str and type_.startswith(("int", "uint")):
+        if value[:2] == "0x":
+            return (type_, to_int(hexstr=value))
+        else:
+            return (type_, to_int(text=value))
 
     return (type_, value)
 
@@ -125,7 +116,7 @@ def find_type_dependencies(type_, types, results=None):
             f"{type_} of type {type(type_)}"
         )
     # get core type if it's an array type
-    type_ = _parse_core_array_type(type_)
+    type_ = parse_core_array_type(type_)
 
     # don't look for dependencies of solidity types
     if type_ in EIP712_SOLIDITY_TYPES:
@@ -174,8 +165,8 @@ def encode_data(
     types: Dict[str, List[Dict[str, str]]],
     data: Dict[str, Any],
 ) -> bytes:
-    encoded_types = ["bytes32"]
-    encoded_values = [hash_type(type_, types)]
+    encoded_types: List[str] = ["bytes32"]
+    encoded_values: List[Union[bytes, int]] = [hash_type(type_, types)]
 
     for field in types[type_]:
         type, value = encode_field(
@@ -196,8 +187,16 @@ def hash_struct(
     data: Dict[str, Any],
 ) -> bytes:
     encoded = encode_data(type_, types, data)
-    hashed = keccak(encoded)
-    return hashed
+    return keccak(encoded)
+
+
+def hash_EIP712_message(
+    # returns the same hash as `hash_struct`, but automatically determines primary type
+    message_types: Dict[str, List[Dict[str, str]]],
+    message_data: Dict[str, Any],
+) -> bytes:
+    primary_type = get_primary_type(message_types)
+    return keccak(encode_data(primary_type, message_types, message_data))
 
 
 def hash_domain(domain_data: Dict[str, Any]) -> bytes:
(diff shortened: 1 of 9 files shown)


eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,221 @@
...
+    if not isinstance(type_, str):
+        raise ValueError(
+            "Invalid find_type_dependencies input: expected string, got "
+            f"{type_} of type {type(type_)}"
+        )
+    # get core type if it's an array type
+    type_ = parse_core_array_type(type_)
+
+    # don't look for dependencies of solidity types
+    if type_ in EIP712_SOLIDITY_TYPES:
```
fselmo, line 122: tiny nit: two bits of logic with the same return statement can be combined. 

`if...elif` can be used here as well.

```py
if (
    # don't look for dependencies of solidity types
    type_ in EIP712_SOLIDITY_TYPES
    # found a type that's already been added
    or type_ in results
):
    return results

File: eth_account/_utils/encode_typed_data/encoding_and_hashing.py [hunk 2]
```diff
@@ -0,0 +1,221 @@
...
+def hash_struct(
+    type_: str,
+    types: Dict[str, List[Dict[str, str]]],
+    data: Dict[str, Any],
+) -> bytes:
+    encoded = encode_data(type_, types, data)
+    return keccak(encoded)
+
+
+def hash_EIP712_message(
```
fselmo, line 193: nit: [PEP8: N802](https://pypi.org/project/pep8-naming/) wants this to be `hash_eip712_message()`, getting a PyCharm warning.

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,221 @@
...
+    # returns the same hash as `hash_struct`, but automatically determines primary type
+    message_types: Dict[str, List[Dict[str, str]]],
+    message_data: Dict[str, Any],
+) -> bytes:
+    primary_type = get_primary_type(message_types)
+    return keccak(encode_data(primary_type, message_types, message_data))
+
+
+def hash_domain(domain_data: Dict[str, Any]) -> bytes:
+    EIP712_domain_map = {
```
fselmo, line 203: nit: [PEP8: N806](https://pypi.org/project/pep8-naming/) wants this function variable to be lowercase, getting a PyCharm warning as well.

eth_account/_utils/encode_typed_data/helpers.py
```diff
@@ -0,0 +1,61 @@
+from typing import (
+    Any,
+)
+
+
+def _get_EIP712_solidity_types():
```
fselmo, line 6: N802

fselmo (commented):
This PR is great. First pass was strictly just readability and consistency stuff. I'd like to review the logic separate and understand what the test cases even are.

Where did we get all the `expected` hashes for the tests? Is there a decent way to compare with *eth-sig-utils* at the moment without having to rewrite all the same test cases there? How did you go about testing these, did you run the same examples against *ethers.js* / *eth-sig-utils*? Do you have any of that set up somewhere so we don't have to rewrite it?

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,221 @@
...
+    types: Dict[str, List[Dict[str, str]]],
+    name: str,
+    type_: str,
+    value: Any,
+) -> Tuple[str, Union[int, bytes]]:
+    if type_ in types.keys():
+        # type is a custom type
+        if value is None:
+            return (
+                "bytes32",
```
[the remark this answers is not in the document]
pacrob: whoah :exploding_head:

eth_account/messages.py
```diff
@@ -125,6 +132,33 @@ def encode_structured_data(
...
+    warnings.warn(
+        "`encode_structured_data` is deprecated and will be removed in a"
+        " future release. Use encode_typed_data instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return encode_structured_data_legacy(primitive, hexstr=hexstr, text=text)
+
+
+def encode_structured_data_legacy(
```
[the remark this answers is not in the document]
pacrob: Yup, I think 2 half-ideas came together - At first I was thinking to just deprecate the name `encode_structured_data` and have it live on into the future as `..._legacy`, but then the plan changed to just deprecate the whole thing. I'll put the naming back as it was.

reedsa (approved):
Lgtm

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,209 @@
...
+    elif is_array_type(type_):
+        type_ = parse_parent_array_type(type_)
+        type_value_pairs = [encode_field(types, name, type_, item) for item in value]
+        data_types, data_hashes = zip(*type_value_pairs)
+        return ("bytes32", keccak(encode(data_types, data_hashes)))
+
+    elif type_ == "bool":
+        return (type_, coerce_bool(value))
+
+    elif type_ == "bytes":
```
fselmo, line 75: Summarizing what we talked about and trying things on my end, comparing with `eth-sig-util` tests, this should include the following logic in one way or another:

```py
    if is_array_type(type_):
        ...
    elif type_ == "bool":
        ...
    elif type_.startswith("bytes") and not is_array_type(type_):
        if not isinstance(value, bytes):
            value = (
                to_bytes(hexstr=value)
                if is_0x_prefixed_hexstr(value)
                else to_bytes(text=value)
                if isinstance(value, str)
                else to_bytes(value)
            )

        return (
            # keccak hash if dynamic `bytes` type
            ("bytes32", keccak(value))
            if type_ == "bytes"
            # if fixed bytesXX type, do not hash
            else ("bytes32", value)
        )
```

Where the helper method `is_0x_prefixed_hexstr()` checks for the `0x` at the beginning. Otherwise, strings are counted as a hexstring if they pass the regex [here](https://github.com/ethereum/eth-utils/blob/e1a395207f26962e49147fc80ccff23f150b2d62/eth_utils/hexadecimal.py#L11-L11), which doesn't require the `0x`. This means a string that is meant to be a string representation of the number "10" is treated as a hex string instead. "0x10" is very different from "10", which is why those tests were failing.

On my end, I included the `is_0x_prefixed_hexstr()` method in your `encode_typed_data/helpers.py` file and it just looks like this:

```py
def is_0x_prefixed_hexstr(value: Any):
    return is_hexstr(value) and value.startswith("0x")
```

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,209 @@
...
+    elif type_ == "string":
+        if isinstance(value, int):
+            value = to_bytes(value)
+        else:
+            value = to_bytes(text=value)
+        return ("bytes32", keccak(value))
+
+    # allow string values for int and uint types
+    elif type(value) == str and type_.startswith(("int", "uint")):
+        if value[:2] == "0x":
```
fselmo, line 91: nit: This could also just use `is_0x_prefixed_hexstr()` from the above comment for consistency

eth_account/_utils/encode_typed_data/helpers.py
```diff
@@ -0,0 +1,61 @@
+from typing import (
+    Any,
+)
+
+
+def _get_eip712_solidity_types():
+    types = ["bool", "address", "string", "bytes"]
```
fselmo, line 7: It seems that `V4` allows for `uint` and `int` types and `eth-sig-util` tests for these. Just by adding the types to this list, our implementation in this PR seems to pass all of those tests.

```suggestion
    types = ["bool", "address", "string", "bytes", "uint", "int"]
```

fselmo (commented):
Adding a few more comments inline, some which we talked about, and some that I ended up debugging to find the issue. See the `is_0x_prefixed_hexstr()` helper method for why some of the `bytes` tests were failing when comparing to `eth-sig-util`.

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,209 @@
...
+            return (type_, to_int(hexstr=value))
+        else:
+            return (type_, to_int(text=value))
+
```
fselmo, line 95: I'm not sure on this one tbh... but if we want to be a bit more forgiving with `address` type as well, we could add something like:

```py
    elif type_ == "address":
        if is_0x_prefixed_hexstr(value):
            value = to_bytes(hexstr=value)
        elif isinstance(value, int):
            value = to_bytes(value)

        if len(value) < 20:
            # pad values that are not 20 bytes
            value = value.rjust(20, b"\x00")

        return (type_, value)
```

This would pass some of the `eth-sig-util` test cases such as encoding type `address` with value:

- `0x0`
- `0x1fffffffffffff`
- `0x10`
- `10`

etc... which `eth-sig-util` tests against

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -0,0 +1,209 @@
...
+        raise ValueError(f"Missing value for field `{name}` of type `{type_}`")
+
+    elif is_array_type(type_):
+        type_ = parse_parent_array_type(type_)
+        type_value_pairs = [encode_field(types, name, type_, item) for item in value]
+        data_types, data_hashes = zip(*type_value_pairs)
+        return ("bytes32", keccak(encode(data_types, data_hashes)))
+
+    elif type_ == "bool":
+        return (type_, coerce_bool(value))
```
fselmo, line 73: `bool(value)` seems to be consistent with everything that `eth-sig-util` is expecting. Are there examples where the logic within `coerce_bool()` is expected?

fselmo: Found some cases, actually. For python bytes types, the bool(value) will treat all bytes as "true", whereas javascript / eth-sig-utils will evalutate the value as a number, rendering 0 as False and any other byte value as True. We could likely use if isinstance(value, int) -> bool(to_bytes(value)).

tests/core/test_encode_typed_data.py
```diff
@@ -0,0 +1,1024 @@
...
+                "expected_exception": ValueError,
+                "match": "Invalid domain key: `classification`",
+            },
+        ),
+    ),
+    ids=[
+        "key in domain_data not in EIP712 domain fields",
+    ],
+)
+def test_hash_domain_fail(domain_data, expected):
```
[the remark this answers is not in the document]
pacrob: Both it and the ..._pass could use more cases anyway. And by using the `ids` list, it gives you the name nicely in the output when the test fails.  I'll put a few more edgy cases in.

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
pacrob: good stuff, added

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
pacrob: I feel like address should be less forgiving in this regard. The 0x-prefixed string vs non-prefixed I can get behind, but it really seems like if a message has come through with an address of "0x10" that a mistake has been made somewhere.

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
pacrob: Nope, `coerce_bool` was my own "reasonableness" addition. `bool(value)` works for me.

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
fselmo: I agree with you there though. I'm OK not adding any of this, even non-0x prefixed. It seems to give different results in `eth-sig-util`. We can leave this out entirely, just figured I'd put it here for discussion.

eth_account/_utils/encode_typed_data/helpers.py
pacrob: added

tests/core/test_encode_typed_data.py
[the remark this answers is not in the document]
pacrob: added one more

fselmo (approved):
Just a small nit in capitalization with the last commit. Nothing blocking. This looks great, thanks for all the work here 👍🏼 👍🏼

PR #236, opened 2023-08-28
update ethers integration testing

pacrob:
### What was wrong?

Ethers mnemonic/hd wallet testing was using an old version of ethers. More javascript testing will be added with incoming typed-data signing, so upgrading this now in preparation.

### How was it fixed?

Upgraded `ethers` version to the latest, and bumped `node.js` version from 18 to 20.

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/tests/integration/ethers-cli/setup_node_v20.sh
+++ b/tests/integration/ethers-cli/setup_node_v20.sh
@@ -3,14 +3,14 @@
 # Discussion, issues and change requests at:
 #   https://github.com/nodesource/distributions
 #
-# Script to install the NodeSource Node.js 18.x repo onto a
+# Script to install the NodeSource Node.js 20.x repo onto a
 # Debian or Ubuntu system.
 #
 # Run as root or insert `sudo -E` before `bash`:
 #
-# curl -sL https://deb.nodesource.com/setup_18.x | bash -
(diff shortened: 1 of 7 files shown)


tests/integration/ethers-cli/setup_node_v20.sh
```diff
@@ -3,14 +3,14 @@
 # Discussion, issues and change requests at:
 #   https://github.com/nodesource/distributions
 #
-# Script to install the NodeSource Node.js 18.x repo onto a
+# Script to install the NodeSource Node.js 19.x repo onto a
```
fselmo, line 6: Should these be `20.x`?

fselmo (approved):
Good stuff. lgtm. Just a small comment, non-blocking.

tests/integration/ethers-cli/setup_node_v20.sh
pacrob: It's the script that pulled from `https://deb.nodesource.com/setup_20.x`, but it does look like they forgot to update that part

pacrob: updated for make better

PR #238, opened 2023-09-08
Add ability to call encode_typed_data with a single dict argument

pacrob:
### What was wrong?

The former `encode_structured_data`, Mask, and even the original EIP expect the message to be a single json object/python dict. 

### How was it fixed?

 - Added the different arg option, plus docs and an example. 
 - Added tests to make sure calling `encode_typed_data` with 3 args provides the same result as with 1 arg
 - Found a bug in handling empty and missing arrays in message, fixed and added tests

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/_utils/encode_typed_data/encoding_and_hashing.py
+++ b/eth_account/_utils/encode_typed_data/encoding_and_hashing.py
@@ -63,8 +63,24 @@ def encode_field(
         raise ValueError(f"Missing value for field `{name}` of type `{type_}`")
 
     elif is_array_type(type_):
-        type_ = parse_parent_array_type(type_)
-        type_value_pairs = [encode_field(types, name, type_, item) for item in value]
+        # handle array type with non-array value
+        if not isinstance(value, list):
+            raise ValueError(
+                f"Invalid value for field `{name}` of type `{type_}`: "
+                f"expected array, got `{value}` of type `{type(value)}`"
+            )
+
+        parsed_type = parse_parent_array_type(type_)
+        type_value_pairs = [
+            encode_field(types, name, parsed_type, item) for item in value
+        ]
+        if not type_value_pairs:
+            # the keccak hash of `encode((), ())`
+            return (
+                "bytes32",
+                b"\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p",  # noqa: E501
+            )
+
         data_types, data_hashes = zip(*type_value_pairs)
         return ("bytes32", keccak(encode(data_types, data_hashes)))
 
--- a/eth_account/messages.py
+++ b/eth_account/messages.py
@@ -26,6 +26,7 @@ from hexbytes import (
 )
 
 from eth_account._utils.encode_typed_data.encoding_and_hashing import (
+    get_primary_type,
     hash_domain,
     hash_eip712_message,
 )
@@ -346,26 +347,34 @@ def defunct_hash_message(
 
 
 def encode_typed_data(
-    domain_data: Dict[Any, Any],
-    message_types: Dict[Any, Any],
-    message_data: Dict,
+    domain_data: Dict[str, Any] = None,
+    message_types: Dict[str, Any] = None,
+    message_data: Dict[str, Any] = None,
+    full_message: Dict[str, Any] = None,
 ) -> SignableMessage:
     r"""
     Encode an EIP-712_ message in a manner compatible with other implementations
     in use, such as the Metamask and Ethers ``signTypedData`` functions.
 
-    Supply the message as exactly three arguments:
+    You may supply the information to be encoded in one of two ways:
+
+    As exactly three arguments:
 
         - ``domain_data``, a dict of the EIP-712 domain data
-        - ``message_types``, a dict of custom types
+        - ``message_types``, a dict of custom types (do not include a ``EIP712Domain``
+          key)
         - ``message_data``, a dict of the data to be signed
+
+    Or as a single argument:
+
+        - ``full_message``, a dict containing the following keys:
+            - ``types``, a dict of custom types (may include a ``EIP712Domain`` key)
+            - ``primaryType``, (optional) a string of the primary type of the message
+            - ``domain``, a dict of the EIP-712 domain data
+            - ``message``, a dict of the data to be signed
 
     .. WARNING:: Note that this code has not gone through an external audit, and
         the test cases are incomplete.
-
-    Usage Notes:
-        - ``message_types`` should not include the ``EIP712Domain`` key. It will be
-            derived from ``domain_data``.
 
     Type Coercion:
         - For fixed-size bytes types, smaller values will be padded to fit in larger
@@ -385,15 +394,17 @@ def encode_typed_data(
     :param domain_data: EIP712 domain data
     :param message_types: custom types used by the `value` data
     :param message_data: data to be signed
+    :param full_message: a dict containing all data and types
     :returns: a ``SignableMessage``, an encoded message ready to be signed
 
 
     .. doctest:: python
 
-        >>> # an example of basic usage
+        >>> # examples of basic usage
         >>> import json
         >>> from eth_account import Account
         >>> from eth_account.messages import encode_typed_data
+        >>> # 3-argument usage
 
         >>> # all domain properties are optional
         >>> domain_data = {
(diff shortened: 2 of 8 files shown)


eth_account/messages.py
```diff
@@ -432,10 +443,107 @@ def encode_typed_data(
...
+        # If EIP712Domain types were provided, check that they match the domain data
+        if "EIP712Domain" in full_message["types"]:
+            domain_data_keys = full_message["domain"].keys()
+            domain_types_keys = [
+                field["name"] for field in full_message["types"]["EIP712Domain"]
+            ]
+
+            if set(domain_data_keys) != (set(domain_types_keys)):
+                raise ValidationError(
+                    "The fields provided in ``domain`` do not match the fields provided"
```
reedsa, line 515: I haven't seen a lot of messages raised with any readable formatting, it's usually just a long string. Curious if some newlines in the exceptions would make it easier to line up the two sets for better readability.

eth_account/messages.py
```diff
@@ -346,27 +347,35 @@ def defunct_hash_message(
 
 
 def encode_typed_data(
-    domain_data: Dict[Any, Any],
-    message_types: Dict[Any, Any],
-    message_data: Dict,
+    domain_data: Dict[Any, Any] = None,
```
fselmo, line 350: Is this ever not `Dict[str, Any]`?

eth_account/_utils/encode_typed_data/encoding_and_hashing.py
```diff
@@ -63,9 +63,21 @@ def encode_field(
...
+            )
+
+        parsed_type = parse_parent_array_type(type_)
+        type_value_pairs = [
+            encode_field(types, name, parsed_type, item) for item in value
+        ]
+        if type_value_pairs:
+            data_types, data_hashes = zip(*type_value_pairs)
+        else:
+            data_types, data_hashes = (), ()
```
fselmo, line 80: It seems like this is always a constant (`(), ()`). We should return that constant here instead and forego the need to `keccak(encode(empty_tuples))` every time this conditional is true. We can add a comment in the code to clarify what that constant means.

```py
if not type_value_pairs:
    # the keccak hash of encode(`(), ()` )
    return ("bytes32", constant_value_here)
    
data_types, data_hashes = zip(*type_value_pairs)

return return ("bytes32", keccak(encode(data_types, data_hashes)))

**Review by fselmo — COMMENTED** [review]
Mostly just nits, nothing blocking. I strongly feel like we should combine these single json file test cases, as that's pretty easy to do with JSON. These tests could be boiled down into three files with good comments on the test cases if need be:

`tests/eip712_messages/invalid.json`
`tests/eip712_messages/invalid_one_arg.json`  maybe could be in invalid too? but this is fine
`tests/eip712_messages/valid.json`

each of them looks like

```json
[
    // bad uint field type in types passed a string
    {...},
    // bad uint field type in types passed an int
    {...},
    {...},
    {...},
]
```

This could help us encourage even more test cases by not bloating up a directory with single test cases if they fit into the same category of tests.

tests/core/test_encode_typed_data/test_encode_typed_data_full_messages.py
```diff
@@ -0,0 +1,69 @@
...
+
+    with pytest.raises(ValueError):
+        encode_typed_data(*convert_to_3_arg(loaded_files[message]))
+
+
+@pytest.mark.parametrize("message", one_arg_invalid)
+def test_messages_that_are_only_invalid_for_one_arg_encoding(message):
+    with pytest.raises(ValidationError):
+        encode_typed_data(full_message=loaded_files[message])
+    try:
```
fselmo, line 66: Curious why the `try` / `except` here? If the call without `full_message` fails, it will fail the test. Maybe leave an inline comment in the test if we want to clarify why it shouldn't throw an exception.

eth_account/messages.py
pacrob: Noop, fixing

eth_account/messages.py
```diff
@@ -346,27 +347,35 @@ def defunct_hash_message(
...
+    message_types: Dict[Any, Any] = None,
+    message_data: Dict = None,
```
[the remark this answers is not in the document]
pacrob: Yep! fixing

eth_account/messages.py
```diff
@@ -432,10 +443,107 @@ def encode_typed_data(
...
+        ...         },
+        ...         "to": {
+        ...             "name": "Bob",
+        ...             "wallet": "0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB"
+        ...         },
+        ...         "contents": "Hello, Bob!",
+        ...     },
+        ... }
+        >>> signable_msg_2 = encode_typed_data(full_message=full_message)
+        >>> signed_msg_2 = Account.sign_message(signable_msg, key)
```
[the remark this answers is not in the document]
pacrob: yep, thanks!

eth_account/messages.py
```diff
@@ -432,10 +443,107 @@ def encode_typed_data(
...
+        >>> signed_msg_2.messageHash
+        HexBytes('0xc5bb16ccc59ae9a3ad1cb8343d4e3351f057c994a97656e1aff8c134e56f7530')
+        >>> signed_msg_2 == signed_msg
+        True
     .. _EIP-712: https://eips.ethereum.org/EIPS/eip-712
     """
+    if full_message is not None and (
+        domain_data is not None or message_types is not None or message_data is not None
```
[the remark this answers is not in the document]
pacrob: yep, much nicer, thanks!

eth_account/messages.py
```diff
@@ -432,10 +443,107 @@ def encode_typed_data(
...
+    ):
+        raise ValueError(
+            "You may supply either `full_message` or `domain_data`, `message_types`, "
```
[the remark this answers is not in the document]
pacrob: Updated for a bit more clarity. OK?

eth_account/messages.py
pacrob: This is going to refer only to the keys within the domain section of the message, of which there are only 5 valid options. The full error message will be something like: 

```py
ValidationError: The fields provided in `domain` do not match the fields provided in `types.EIP712Domain`. The fields provided in `domain` were `['name', 'version', 'chainId', 'verifyingContract', 'salt']`, but the fields provided in `types.EIP712Domain` were `['name', 'version', 'chainId', 'verifyingContract']`.
```

I do like the idea of newlines to aid comparison, but it seems overkill for comparing lists of 5 or fewer items. Leaving as-is for now, but happy to discuss further!

eth_account/messages.py
```diff
@@ -432,10 +443,107 @@ def encode_typed_data(
...
+
+            if set(domain_data_keys) != (set(domain_types_keys)):
+                raise ValidationError(
+                    "The fields provided in ``domain`` do not match the fields provided"
+                    " in ``types.EIP712Domain``. The fields provided in ``domain`` were"
+                    f" {domain_data_keys}, but the fields provided in "
+                    f"``types.EIP712Domain`` were {domain_types_keys}."
+                )
+
+        types_less_domain = full_message["types"].copy()
```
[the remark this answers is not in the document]
pacrob: sounds good, changed

eth_account/messages.py
```diff
@@ -432,10 +443,107 @@ def encode_typed_data(
...
+            provided_primary_type = full_message["primaryType"]
+            if derived_primary_type != provided_primary_type:
+                raise ValidationError(
+                    "The provided `primaryType` does not match the derived "
+                    "`primaryType`. The provided `primaryType` was "
+                    f"{provided_primary_type}, but the derived `primaryType` was "
+                    f"{derived_primary_type}."
+                )
+
+        parsed_domain_data = full_message["domain"]
```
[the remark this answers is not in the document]
pacrob: changed

tests/core/test_encode_typed_data/test_encode_typed_data_full_messages.py
pacrob: Noop, just being a little overly-explicit I think. changed!

pacrob:
> I strongly feel like we should combine these single json file test cases

Done! And I learned that pytest will easily parametrize a dict and use the keys for the test id.

reedsa (approved):
I do like how the assertions use the parameterized variable directly. Nice work! 💯 

If we end up doing something similar later or adding more permutations of messages, we might consider finding a clever way to use parameterized fixtures for the data.

fselmo (approved):
Love it. Nice work 👌🏼

PR #239, opened 2023-09-26
Add `sign_typed_data`

pacrob:
### What was wrong?

`sign_typed_data` is not accessible via the `Account` class.

### How was it fixed?

Added it.
Condensed some line-too-long ignores (`# noqa: E501`) so they don't appear in the docs

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/test_accounts.py
+++ b/tests/core/test_accounts.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+import json
 import os
 
 from cytoolz import (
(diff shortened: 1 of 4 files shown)


tests/core/test_accounts.py
```diff
@@ -818,3 +823,107 @@ def test_eth_account_prepared_encrypt(
...
+                    },
+                    {
+                        "name": "Dot",
+                        "wallet": "0xdddddddddddddddddddddddddddddddddddddddd",
+                    },
+                ],
+                "contents": "Hello, Bob!",
+            },
+            HexBytes(
+                "0xfcd28bfd425a26215d645481af446895f761b16896ac007c86433136ad9bb5c815ccdef9a6947396be54e92955c5c707ce65099d0371b4488143264675f5e18b1c"  # noqa: E501
```
fselmo, line 916: How do we tell if this is right or wrong? 🤔

fselmo (approved):
lgtm based on the other tests.

 Just a q on the `test_sign_typed_data_produces_expected_hash()` test. Where does the "expected_sig" hash come from? How do we come to expect it?

tests/core/test_accounts.py
pacrob: I signed the message with ethers.js, eth-sig-utils, encode_structured_data+sign, and sign_typed_data and got the same sig from all of them.

PR #240, opened 2023-10-16
add upper pin to hexbytes dep due to incoming breaking update in 0.4.0

pacrob:
### What was wrong?

The hexbytes lib will be changing the way Hexbytes.hex() works - https://github.com/ethereum/hexbytes/issues/14

### How was it fixed?

Add upper pin to hexbytes dep now, then after next release use hexbytes 0.4.0 as a lower bound.

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
         "eth-keys>=0.4.0",
         "eth-rlp>=0.3.0",
         "eth-utils>=2.0.0",
-        "hexbytes>=0.1.0",
+        "hexbytes>=0.1.0,<0.4.0",
         "rlp>=1.0.0",
     ],
     python_requires=">=3.7, <4",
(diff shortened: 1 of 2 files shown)


reedsa (approved):

setup.py
```diff
@@ -66,7 +66,7 @@
         "eth-keys>=0.4.0",
         "eth-rlp>=0.3.0",
         "eth-utils>=2.0.0",
-        "hexbytes>=0.1.0",
+        "hexbytes>=0.1.0,<=0.3.1",
```
fselmo, line 69: If we plan on making the breaking change in `0.4.0`, we should use `<0.4.0` here. That way if we want to backport any changes to `0.3.x`, we can. Thoughts?

fselmo (approved):
lgtm

PR #241, opened 2023-10-17
Js sign typed data comparison testing

pacrob:
### What was wrong?

`sign_typed_data` has been implemented with the intention of producing similar results as metamask and ethers-js, but the current test cases only compare with a static value (hash or sig)

### How was it fixed?

 - Built on the existing javascript mnemonic testing, verify messages encode to the same sig where expected. Test cases explain why where they don't match.
 - Fixed a bug in encoding where the order of encoding the domain data was determined by the data (a json object, which is unordered) instead of the standard domain types (an array, which is ordered).

### Todo:
- [ ] Clean up commit history

#### Cute Animal Picture

image


.circleci/config.yml
```diff
@@ -88,10 +88,10 @@ integration: &integration
           python -m pip install tox
     - run:
         name: install dependencies (node)
-        command: sudo -E bash tests/integration/ethers-cli/setup_node_v20.sh && sudo apt-get install -y nodejs
+        command: sudo -E bash tests/integration/js-integration-test-scripts/setup_node_v20.sh && sudo apt-get install -y nodejs
     - run:
-        name: Build ethers-cli
-        command: cd tests/integration/ethers-cli && sudo npm install && sudo npm install -g . && cd ../../../
+        name: Build js-integration-test-scripts
```
reedsa, line 93: This name seems redundant since it is located in the integration directory. I don't have context about the ethers tests or what that name was about. Do the tests encompass tests for any js client? Would be curious if there is a name that aligns with what the tests do?

reedsa (commented):
Nit over the naming, but overall lgtm!

tests/eip712_messages/valid_py_and_ethers.json
```diff
@@ -0,0 +1,220 @@
+{
+    "valid_space_in_custom_type_metamask_returns_empty_string_if_non_alphanum_in_custom_type": {
```
fselmo, line 2: More of a question. Is this something we _want_ to allow? Is there a benefit I'm not seeing? It seems to me like spaces aren't necessarily _meant_ to be in these definitions. Does `ethers.js` go out of its way to make sure that this is valid?

Curious on the reasoning here as to why we'd want this in `py`.

fselmo (commented):
First pass looks great to me 🔥

I had a question on one of the validations we do that matches *ethers.js* but not *metamask*. I'm going to run the tests now and make sure I understand how they are set up on the JS side as my JS is rusty 😅 ... but the rest looks good to me 👍🏼

tests/eip712_messages/valid_py_and_ethers.json
pacrob: Source of inspiration for the test is here: https://ethereum.stackexchange.com/questions/153221/how-to-sign-message-eip-712-with-whitespace-in-name

EIP712 says a type should be a "valid identifier", which in solidity-speak is only _, &, and alphanumerics. 

Ethers pays no attention to this - if it's a valid string in JS, it encodes it:
https://github.com/ethers-io/ethers.js/blob/e454afb2fa3612c273a09b154c36b35fafab17b1/src.ts/hash/typed-data.ts#L172

Metamask uses a faulty regex in a possible attempt to match the "valid identifier". Faulty, because if it finds a non-alphanumeric, instead of throwing an error it just returns an empty string.
https://github.com/MetaMask/eth-sig-util/blob/10206bf2f16f0b47b1f2da9a9cfbb39c6a7a7800/src/sign-typed-data.ts#L434

Given that these messages are JSON-formatted, it seems to me most helpful to the user to allow any valid JSON string in an identifier.

fselmo: > Faulty, because if it finds a non-alphanumeric, instead of throwing an error it just returns (and then encodes) an empty string.

Yikes, yeah. Fair enough. Thanks for the details 👍🏼

fselmo (approved):
lgtm 👍🏼

.circleci/config.yml
pacrob: Ethers is to JS as web3py is to python - https://github.com/ethers-io/ethers.js/tree/e454afb2fa3612c273a09b154c36b35fafab17b1

Prior to this PR, the integration tests there only used a specific set of functions from ethers, so the name was appropriate. Since I've added tests that use both ethers and a metamask library, I figured a more general name was in order.

pacrob: Agree on the redundancy - I'll clean that up.

pacrob: Okey, shortened to `js-scripts`

PR #244, opened 2023-11-30
remove deprecated camelCase methods that have a direct snake replacement

pacrob:
### What was wrong?

Several methods that were originally camelCase had direct snake_case replacements. They were deprecated ~5 years ago.

### How was it fixed?

Removed the deprecated camelCase version.

Does not touch the `signHash` -> `sign_message` deprecation, as there's some discussion there sort out.

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/signers/base.py
+++ b/eth_account/signers/base.py
@@ -63,23 +63,6 @@ class BaseAccount(ABC):
         pass
 
     @abstractmethod
-    def signTransaction(self, transaction_dict):
-        """
-        Sign a transaction dict.
-
-        This uses the same structure as in
-        :meth:`~eth_account.account.Account.sign_transaction`
-        but without specifying the private key.
-
-        .. CAUTION:: Deprecated for
-            :meth:`~eth_account.account.signers.local.sign_transaction`.
-            This method will be removed in v0.6
-
-        :param dict transaction_dict: transaction with all fields specified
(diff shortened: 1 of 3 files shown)


pacrob:
> Are we ready for a major release without backporting?

Yes, we can release all we want, major or minor, as long as we keep the `hexbytes<0.4.0`. Once web3 v7 is out (which will include `hexbytes>=1.0.0`), we will go through all the other libs that depend on hexbytes, change them to `>=1.0.0`, and release major.

fselmo:
If we don't have upper pins on depending libraries and they haven't
switched to snake case, then that would break, no?

pacrob:
> If we don't have upper pins for eth-account on depending libraries and they haven't switched to snake case, then that would break, no?

Yes, they would. That's the danger of not having upper pins.

reedsa (approved):
Lgtm

kclowes (approved):
Just left a comment around the `signHash` documentation. If there is documentation elsewhere around `signHash`, then LGTM! Otherwise we should maybe keep it, since `signHash` is still a thing if I understand your comment correctly. 

I looked and I don't think web3 is using `signTransaction` anywhere so we should be good on that front. We are still using `signHash` though.

PR #253, opened 2024-03-19
Cancun updates

fselmo:
### What was wrong?

- Add support for type `3` blob transactions
- Refactor typed transactions into separate files for better organization
- Normalize TypedTransaction.as_dict() so that tupleized or list-based `accessList` or `blobVersionedHashes` within a dict all normalize to tuples for consistency and ease of comparison.

### Todo:

- [x] Implement a new API to wrap blob txs with the appropriate blobs, commitments, and proofs
- [x] Add tests for the wrapped `PooledTransaction` API

#### Cute Animal Picture

Screenshot 2024-03-19 at 09 53 18


pacrob (approved):
lgtm! Nice reorg

eth_account/_utils/typed_transactions/base.py
```diff
@@ -0,0 +1,78 @@
...
+        ),
+        "maxPriorityFeePerGas": hexstr_if_str(to_int),
+        "maxFeePerGas": hexstr_if_str(to_int),
+        "maxFeePerBlobGas": hexstr_if_str(to_int),
+        "blobVersionedHashes": apply_formatter_to_array(hexstr_if_str(to_bytes)),
+    },
+)
+
+
+class _TypedTransactionImplementation(ABC):
```
kclowes, line 239: Should `assert_valid_fields`, `from_dict` or `from_bytes` be in here at all?

tests/core/test_typed_transactions.py
```diff
@@ -3,9 +3,16 @@
 )
 import pytest
 
-from eth_account._utils.typed_transactions import (
+from eth_account._utils.typed_transactions.access_list_transaction import (
```
kclowes: nit: you should be able to import these from `eth_account._utils.typed_transactions` still with the `__init__.py`, I believe

eth_account/_utils/typed_transactions/blob_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+        if not all(valid_fields.values()):
+            invalid = {
+                key: dictionary[key] for key, valid in valid_fields.items() if not valid
+            }
+            raise TypeError(f"Transaction had invalid fields: {repr(invalid)}")
+
+    @classmethod
+    def from_dict(cls, dictionary: Dict[str, Any]) -> "BlobTransaction":
+        """
+        Builds a DynamicFeeTransaction from a dictionary.
```
kclowes:
```suggestion
        Builds a BlobTransaction from a dictionary.
```

eth_account/_utils/typed_transactions/blob_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+                f"expected transaction type {cls.transaction_type}, "
+                f"got {transaction_type}"
+            )
+        return cls(
+            dictionary=sanitized_dictionary,
+        )
+
+    @classmethod
+    def from_bytes(cls, encoded_transaction: HexBytes) -> "BlobTransaction":
+        """Builds a DynamicFeeTransaction from a signed encoded transaction."""
```
kclowes:
```suggestion
        """Builds a BlobTransaction from a signed encoded transaction."""
```

kclowes (approved):
Looks good to me. Just left a few nits

fselmo:
New API for generating commitments and proofs from blobs added with pydantic models for validation. RLP serialization and deserialization with the `PooledTransaction` payload from EIP-4844 added. I want to add more tests but asking for a re-review so we can chat about the API.

pacrob (approved):
Apart from the comment on fields to add to `_TypedTransactionImplementation`, lgtm!

eth_account/_utils/typed_transactions/base.py
fselmo: I think it makes sense to have `assert_valid_fields` there but `from_dict` and `from_bytes` return instances specific to the class that they are defined within so I think the signature differences muddy the abstract method + implementation pattern.

fselmo: I could see it going either way... `assert_valid_fields` isn't implemented at the `TypedTransaction` level so I don't think it's a necessity, more of a safety check. And it has a different signature for `BlobTransaction` (`has_blobs: bool = false`) that we'd have to apply across the board since we don't validate certain fields of the dictionary if blobs are also provided.

Curious on thoughts here @pacrob @kclowes

kclowes: Okay, yeah, I think it makes sense to exclude them then. Thanks for the clarification!

PR #258, opened 2024-04-01
Update documentation for blob transaction signing

fselmo:
### What was wrong?

- Documentation was missing for blob transaction signing
- Related to #253

### How was it fixed?

- Update documentation for signing transactions with blob transactions
- Update the other transaction type examples to be more accurate

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2024-04-01 at 15 45 36

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -657,132 +657,114 @@ class Account:
         :param blobs: optional list of blobs to sign in addition to the transaction
         :returns: Various details about the signature - most
           importantly the fields: v, r, and s
-        :rtype: AttributeDict
+        :rtype: SignedTransaction
 
-        .. code-block:: python
+        .. doctest:: python
 
             >>> # EIP-1559 dynamic fee transaction (more efficient and preferred over legacy txn)
+            >>> from eth_account import Account
             >>> dynamic_fee_transaction = {
-                    "type": 2,  # optional - can be implicitly determined based on max fee params
-                    "gas": 100000,
-                    "maxFeePerGas": 2000000000,
-                    "maxPriorityFeePerGas": 2000000000,
-                    "data": "0x616263646566",
-                    "nonce": 34,
-                    "to": "0x09616C3d61b3331fc4109a9E41a8BDB7d9776609",
-                    "value": "0x5af3107a4000",
-                    "accessList": (  # optional
-                        {
-                            "address": "0x0000000000000000000000000000000000000001",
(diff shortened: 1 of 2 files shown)


eth_account/account.py
```diff
@@ -731,16 +733,56 @@ def sign_transaction(self, transaction_dict, private_key, blobs=None):
...
+            >>> key = '0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318'
+            >>> signed = Account.sign_transaction(blob_transaction, key, blobs=[empty_blob])
+            SignedTransaction(
+                rawTransaction=HexBytes('0x03fa020147f8d98205392284773594008477359400830186a09409616c3d61b3331fc4109a9e41a8bdb7d97766098...'),
+                hash=HexBytes('0xf9dc8867c4324fd7f4506622aa700989562770f01d7d681cef74a1a1deb9fea9'),
+                r=14319949980593194209648175507603206696573324965145502821772573913457715875718,
+                s=9129184742597516615341309773045281461399831333162885393648678700392065987233,
+                v=1
+            )
         """  # noqa: E501
```
kclowes, line 768: nit: 
```suggestion
        """
```

eth_account/account.py
```diff
@@ -731,16 +733,56 @@ def sign_transaction(self, transaction_dict, private_key, blobs=None):
...
+                            "storageKeys": (
+                                "0x0100000000000000000000000000000000000000000000000000000000000000",
+                            )
+                        },
+                    ),
+                    "chainId": 1337,
+
+                    # Note: unless you've computed `blobVersionedHashes` yourself and it matches the blob data you are
+                    # sending along with the tx, you should omit the  field since it will be calculated from the blobs
+                    # "blobVersionedHashes": [...]  # optional (but must match the expected value from the blob data)
```
kclowes:
```suggestion
                    "blobVersionedHashes": [...]  # optional (but must match the expected value from the blob data)
```

kclowes (approved):
Similar comments to the `0.11.x` branch, but LGTM!

eth_account/account.py
fselmo: I think this is necessary at the end of the docstring so that all lines in the docstring that break `E501` are accounted for. Unless I missed something but we can probably address it elsewhere if you found a way around that.

PR #260, opened 2024-04-03
Remove signHash in favor of unsafe_sign_hash

### LINKED ISSUE #85: signHash is deprecated in favor of sign_message
After I updated to the last version of `eth-account` I see this warning on my tests. At first I thought it would be just renaming the function, but I realized I cannot sign a custom hash anymore. I understand this can be a way of preventing users to sign hashes of transactions by mistake, but I think a developer should have the possibility of signing hashes and in my case I need it for some contracts that rely on signing hashes and not messages.

## How can it be fixed?

I would propose not deprecating the old method or adding a raw `SignableMessage` that allows to sign hashes.

It could be possible that there's a way to do it and I didn't figure it out, sorry if that's the case

kclowes:
### What was wrong?
It's not safe to sign random hashes, but occasionally developers want to do so. We were undecided about what to do with the method, but decided to add a new method called `unsafe_sign_hash` and finally remove `signHash`

Closes #85

### How was it fixed?
Removed `signHash`, moved functionality to `unsafe_sign_hash`.

LMK if you think we should add more/different docs

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -12,7 +12,6 @@ from typing import (
     Union,
     cast,
 )
-import warnings
 
 from eth_keyfile import (
     create_keyfile_json,
@@ -582,7 +581,7 @@ class Account:
         return cast(SignedMessage, self._sign_hash(message_hash, private_key))
 
     @combomethod
-    def signHash(self, message_hash, private_key):
+    def unsafe_sign_hash(self, message_hash, private_key):
         """
         Sign the provided hash.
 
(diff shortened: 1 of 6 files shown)


fselmo (commented):
lgtm. I think the name change should point most users to reading the documentation.

pacrob (approved):
lgtm!

reedsa (approved):
lgtm!

Uxio0:
Sounds like a really good solution, thanks @kclowes

Issue #270, opened 2024-04-08
Introducing breaking Pydantic dependency in 0.11.1 and 0.12.1


MHHukiewitz:
### What happened?

One of my projects uses Pydantic < 2.0.0. I'd like to use the latest version of `eth-account` in my project too, but this is now not easily possible, due to the latest addition of Pydantic 2.6 in the versions 0.11.1 and 0.12.1.

While it would be not advisable to revert these releases, I'd like to ask for caution when introducing new, well-known dependencies that could break downstream projects, that require different versions of the requirement. These new dependencies should at least trigger a minor upgrade and should not be introduced as a patch version.

Ideally, I would be able to use 0.12.0, but now have to use <0.11.0, without fixing the required `eth-account` version to 0.12.0.

### Fill this section in if you know how this could or should be fixed

Don't add dependencies in patch (x.x.Y) versions.

### eth-account Version

0.11.1, 0.12.1

### Python Version

3.11

kclowes:
Hey @MHHukiewitz, thanks for reaching out. We understand your frustration, and will definitely be more careful about adding well-known dependencies in the future. We do follow semver, so when libraries are in `0.x`, the patch takes on the meaning that minor would if the library were stable. There are plans to move the library to its first major (stable) version upon on a full audit, which will hopefully happen at some point in the next few months.

To give some background, Cancun blob transaction support was released in 0.12.0 (not 0.12.1) but this also introduced breaking changes to HexBytes and places a lower pin on the HexBytes library at that breaking change version. What that meant is there was no Cancun support for the library where there wasn’t this breaking HexBytes change, in 0.11.0. With 0.11.x likely not getting much more attention after the released 0.12.0, the choice to have the back port of the Cancun network upgrade via support of blob (type=3) transactions seemed like a better alternative than not having it. This is in the case that the Pydantic introduction does not break anything, of course. Your library is unfortunately in that latter category. In this case, pinning the library to 0.11.0 should help you. We understand this is not an ideal scenario for users and as stated above this won't be the case moving forward.

We hear you loud and clear and hopefully there weren't too many bumps in the road for you with this change.

kclowes:
@MHHukiewitz We just released `v0.11.2` without the Pydantic dependency. Hopefully that fixes your issue. `eth-account` will have Pydantic going forward from 0.12.x.

PR #272, opened 2024-04-08
bugfix: revert pydantic dependency introduction

fselmo:
### What was wrong?

- The introduction of ``pydantic`` as a dependency in `0.11.x` was not necessary and it was not appropriate for a patch release (see issue ethereum/eth-account#270). This is simple enough to remove for `0.11.x` and leave it for `0.12.x`.

Related to Issue #270

### How was it fixed?

- Revert the use of `pydantic` as a dependency. This is simple enough to handle by defining `__init__()` methods that validate inputs to the `data` coming into each of the blob class objects.

---
Note: By removing `pydantic`, we don't validate the input's `type` in the data models but this should be consistent with the rest of the library up to `0.11.0`. That's one additional benefit we get from the `pydantic` implementation in `0.12.x`.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2024-04-08 at 13 32 52

### DIFF
--- a/eth_account/_utils/typed_transactions/blob_data_models.py
+++ b/eth_account/_utils/typed_transactions/blob_data_models.py
@@ -20,12 +20,6 @@ from eth_utils import (
 from hexbytes import (
     HexBytes,
 )
-from pydantic import (
-    BaseModel,
-    ConfigDict,
-    computed_field,
-    field_validator,
-)
 
 # import TRUSTED_SETUP from ./kzg_trusted_setup.txt
 TRUSTED_SETUP = os.path.join(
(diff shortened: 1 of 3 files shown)


fselmo:
`bugfix` _and_ `removal` newsfragments to explain the patch as a removal of `pydantic` dependency for `0.11.x`?

kclowes (approved):
LGTM!

kclowes:
> bugfix and removal newsfragments to explain the patch as a removal of pydantic dependency for 0.11.x?

I think bugfix might suffice, but I'll leave it up to your good judgement :)

PR #276, opened 2024-04-25
Publicly expose TypedTransactions

### LINKED ISSUE #262: Expose TypedTransactions Publicly
Currently TypedTransaction classes live under _utils. We want to expose those for general use. This should :tm: just involve moving them and changing the import path as needed.

kclowes:
### What was wrong?
Typed Transactions would be handy as public classes. 

Closes #262 

### How was it fixed?
Moved TypedTransaction classes out of `_utils`. I left a few typed transactions utils in the _utils folder. Let me know if they should be moved. I also double checked that the files within `Manifest.in` do get added to the wheel and tarball. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


eth_account/typed_transactions/access_list_transaction.py
```diff
@@ -227,9 +228,13 @@ def payload(self) -> bytes:
...
         Returns this transaction's payload as bytes.
 
-        Here, the TransactionPayload = rlp([chainId,
-        nonce, gasPrice, gasLimit, to, value, data, accessList,
-        signatureYParity, signatureR, signatureS])
+        Here, the transaction payload is:
+
+        .. code-block:: python
+
+            TransactionPayload = rlp([chainId,
```
fselmo, line 233: The ``rlp`` python module can't be used in this way afaik. This may best be left as part of the docstring like it was. It provides a description of the rlp encoding, as it is usually defined in EIPs, rather than direct python code.

eth_account/typed_transactions/blob_transactions/blob_transaction.py
```diff
@@ -332,10 +330,15 @@ def payload(self) -> bytes:
...
 
-        Here, the TransactionPayload = rlp([chainId,
-        nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit, to, value, data,
-        accessList, maxFeePerBlobGas, blobVersionedHashes, signatureYParity, signatureR,
-        signatureS])
+        Here, the transaction payload looks like:
+
+        .. code-block:: python
+
+            TransactionPayload = rlp([chainId,
```
fselmo, line 335: Same as the previous comment here. This could be left in the docstring I think since ``rlp`` doesn't work exactly like this.

eth_account/typed_transactions/dynamic_fee_transaction.py
```diff
@@ -219,9 +219,13 @@ def payload(self) -> bytes:
...
         Returns this transaction's payload as bytes.
 
-        Here, the TransactionPayload = rlp([chainId,
-        nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit, to, value, data,
-        accessList, signatureYParity, signatureR, signatureS])
+        Here, the transaction payload is:
+
+        .. code-block:: python
+
+            TransactionPayload = rlp([chainId,
```
fselmo, line 224: Same comment as the others here.

fselmo (commented):
I don't think the `rlp` module we use in python works the way the docstrings state in this PR. We should probably keep those conceptual and in the docstring but I'm open to any other ideas there. Other than that, this lgtm 👍🏼

eth_account/typed_transactions/access_list_transaction.py
kclowes: Ah, you're right. Changed!

fselmo (approved):
lgtm 👍🏼

PR #280, opened 2024-05-28
add mypy to dev dependencies so it has full context when running, set pre-commit to run it

pacrob:
### What was wrong?

The `mirrors-mypy` pre-commit hook only takes into account the dependencies that are explicitly added in the `.pre-commit-config` file. This misses additional typing context that may come from other dependencies. Specifically here, running mypy via pre-commit passes, but installing `mypy==0.971` and running `mypy -p eth_account` gives:

```
site-packages/pydantic_core/_pydantic_core.pyi:64: error: Positional-only parameters are only supported in Python 3.8 and greater
Found 1 error in 1 file (errors prevented further checking)
```

Bumping to mypy `v1.10.0` in the pre-commit hook yields 25 errors, while installing and running locally yields 32.

### How fixed

Added `mypy==1.10.0` to dev dependencies and set pre-commit to run that instead of its own copy.
Fixed new mypy errors.

Many of the errors are related to `strict_optional` checking. In order to make this PR more focused, I set it to ignore the `strict_optional` rule, can do that in a dedicated PR.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


pyproject.toml
```diff
@@ -24,7 +24,7 @@ disallow_untyped_decorators = false
 disallow_untyped_defs = false
 ignore_missing_imports = true
 strict_equality = true
-strict_optional = true
+strict_optional = false
```
fselmo, line 27: We should create an issue to handle this before merging this PR so we don't lose track of it.

.pre-commit-config.yaml
```diff
@@ -43,11 +43,14 @@ repos:
     -   id: mdformat
         additional_dependencies:
         -   mdformat-gfm
--   repo: https://github.com/pre-commit/mirrors-mypy
```
fselmo, line 46: This is a great change. Admittedly I wasn't looking very close at this and didn't realize we could add steps to run local commands like this. That should be the approach almost every time imo. We should maybe take this time to see if anything else can / should be changed to run a local command instead of some pre-commit repo that _aims_ to do the thing we want it to do.

fselmo (approved):
lgtm, nice catch and thanks for the detailed explanation 👍🏼

.pre-commit-config.yaml
pacrob: I'll go through and verify and file separate PRs if I find anything.

pyproject.toml
pacrob: Made a personal note, issue now added.

.pre-commit-config.yaml
fselmo: For sure, I meant within separate PRs / issues as well. lgtm 👍🏼

kclowes (approved):
LGTM! 🚀

PR #283, opened 2024-06-20
turn on mypy disallow_untyped_defs and disallow_untyped_calls

### LINKED ISSUE #101: Enable stricter mypy checks

### LINKED ISSUE #281: turn `strict_optional` `mypy` flag back on in `pyproject.toml`
### What happened?

It was turned off in PR #280. Turn it back on and fix the errors.

_No response_

### Full error output

_No response_

### Fill this section in if you know how this could or should be fixed

_No response_

### eth-account Version

_No response_

### Python Version

_No response_

### Operating System

_No response_

### Output from `pip freeze`

_No response_

pacrob:
### What was wrong?

Tighten up typing!

### How was it fixed?

Set `disallow_untyped_defs` and `disallow_untyped_calls` to true, fix errors.

Most changes were straightforward. The 2 biggest changes are:
 - `datastructures.py` - much finagling to get `mypy` to accept an overloaded `__getitem__`
 - The files containing `Account` and `LocalAccount` had circular imports once typing was added. Fixed by adding the `AccountLocalActions` ABC for `LocalAccount` to import from. It's not a great name, but the best I could come up with today.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


eth_account/_utils/legacy_transactions.py
```diff
@@ -35,8 +41,34 @@
...
+class UnsignedTransaction(HashableRLP):
+    fields = UNSIGNED_TRANSACTION_FIELDS
+
+
+AnyTransaction = Union[TypedTransaction, Transaction, UnsignedTransaction]
+
 
-def serializable_unsigned_transaction_from_dict(transaction_dict, blobs=None):
+def serializable_unsigned_transaction_from_dict(
+    transaction_dict: Dict[str, Any], blobs: Optional[List[bytes]] = None
```
kclowes: For all of these `transaction_dict`s in this file, can they get tightened up any or should it really be `Dict[str, Any]`?

eth_account/_utils/legacy_transactions.py
```diff
@@ -35,8 +41,34 @@
...
+    ("nonce", big_endian_int),
+    ("gasPrice", big_endian_int),
+    ("gas", big_endian_int),
+    ("to", Binary.fixed_length(20, allow_empty=True)),
+    ("value", big_endian_int),
+    ("data", binary),
+)
+
+
+class Transaction(HashableRLP):
```
kclowes, line 59: Does this make sense to be `SignedTransaction` since it has the `v` `r`, and `s` fields?

eth_account/_utils/legacy_transactions.py
```diff
@@ -61,7 +93,9 @@ def serializable_unsigned_transaction_from_dict(transaction_dict, blobs=None):
     return serializer.from_dict(filled_transaction)
 
 
-def encode_transaction(unsigned_transaction, vrs):
+def encode_transaction(
+    unsigned_transaction: AnyTransaction, vrs: Tuple[int, int, int]
```
kclowes: nit: seems like `unsigned_transaction` should either have a type of `UnsignedTransaction`, or have its name changed to `transaction`, but feel free to ignore that pedantry 😆

eth_account/_utils/legacy_transactions.py
```diff
@@ -141,39 +177,18 @@ def chain_id_to_v(transaction_dict):
...
-        ("r", big_endian_int),
-        ("s", big_endian_int),
-    )
-
-
-class UnsignedTransaction(HashableRLP):
-    fields = UNSIGNED_TRANSACTION_FIELDS
+def fill_transaction_defaults(transaction_dict: Dict[str, Any]) -> Dict[str, Any]:
+    # type ignored because merge does not preserve typing
+    return merge(TRANSACTION_DEFAULTS, transaction_dict)  # type: ignore[no-any-return]
```
kclowes: Can we cast this to `UnsignedTransaction` instead of ignoring?

eth_account/_utils/signing.py
```diff
@@ -27,7 +39,11 @@
 STRUCTURED_DATA_SIGN_VERSION = b"\x01"  # Hex value 0x01
 
 
-def sign_transaction_dict(eth_key, transaction_dict, blobs=None):
+def sign_transaction_dict(
+    eth_key: PrivateKey,
+    transaction_dict: Dict[str, Any],
+    blobs: Optional[List[bytes]] = None,
```
kclowes: Maybe it's worth making a blob type, similar to `Hash32`?

reedsa (approved):
Looks good! I noticed a couple of return statements that show typing issues but I think those can be addressed separately.

eth_account/_utils/validation.py
```diff
@@ -49,14 +55,14 @@ def is_int_or_prefixed_hexstr(val):
...
-def is_empty_or_checksum_address(val):
+def is_empty_or_checksum_address(val: Any) -> bool:
     if val in VALID_EMPTY_ADDRESSES:
         return True
     else:
         return is_valid_address(val)
 
 
-def is_rpc_structured_access_list(val):
+def is_rpc_structured_access_list(val: Sequence[Any]) -> bool:
```
kclowes: I think `val` can be plain old `Any` with the `is_list_like` check on line 67?

eth_account/_utils/validation.py
```diff
@@ -79,7 +85,7 @@ def is_rpc_structured_access_list(val):
     return True
 
 
-def is_rlp_structured_access_list(val):
+def is_rlp_structured_access_list(val: Sequence[Any]) -> bool:
```
kclowes: Same here re. `val` and `Any`

eth_account/hdaccount/mnemonic.py
```diff
@@ -56,7 +56,7 @@
 _cached_wordlists: Dict[str, List[str]] = dict()
 
 
-def get_wordlist(language):
+def get_wordlist(language: str) -> List[str]:
```
kclowes, line 59: I wonder if we should make language an enum instead of accepting any `str`?

eth_account/account_local_actions.py
```diff
@@ -0,0 +1,73 @@
...
+from eth_account.datastructures import (
+    SignedMessage,
+    SignedTransaction,
+)
+from eth_account.messages import (
+    SignableMessage,
+)
+
+
+class AccountLocalActions(ABC):
```
kclowes, line 36: I just noticed that there is already an ABC called BaseAccount. I wonder if we could use that instead of this new ABC? It looks like they're pretty similar.

kclowes (commented):
It will be awesome to get this in! One thing I noticed throughout is that it seems like a lot of values are 32 bytes long (r, s, transaction_hash, etc.)What do you think about making a Bytes32 (maybe a different name) type similar to Hash32 to indicate that this bytes value is exactly 32 bytes? Otherwise, I left some inline comments, let me know what you think!

eth_account/_utils/legacy_transactions.py
pacrob: It does, but then conflicts with the same name in `datastructures.py`. `LegacySignedTransaction` and `LegacyUnsignedTransaction`?

kclowes: Gotcha. Eh, I think all options are about equal, so will leave it up to you! :)

eth_account/_utils/signing.py
pacrob: Done! All `blobs` are now `Blobs` type, imported from a new `types.py`

eth_account/_utils/legacy_transactions.py
pacrob: Is now new `TransactionDictType`

eth_account/account_local_actions.py
pacrob: That's what the `LocalAccount` inherits from. `BaseAccount` and `LocalAccount` are separate entities from `Account` - an instantiated `LocalAccount` has an `Account` as an attribute, and an `Account` can generate new instances of `LocalAccount`. I think it's pretty confusing. But that's why we need something separate.

eth_account/hdaccount/mnemonic.py
pacrob: I think that's a good idea. OK with me doing that in a separate issue/pr?

eth_account/signers/local.py
```diff
@@ -25,34 +57,39 @@ class LocalAccount(BaseAccount):
         b"\\x01\\x23..."
     """
 
-    def __init__(self, key, account):
+    def __init__(self, key: PrivateKey, account: AccountLocalActions):
```
kclowes, line 60: I think this should really just be `Account`?

eth_account/signers/local.py
```diff
@@ -25,34 +57,39 @@ class LocalAccount(BaseAccount):
...
         """
         Initialize a new account with the given private key.
 
         :param eth_keys.PrivateKey key: to prefill in private key execution
         :param ~eth_account.account.Account account: the key-unaware management API
         """
-        self._publicapi = account
+        self._publicapi: AccountLocalActions = account
```
kclowes, line 67: Same here. `AccountLocalActions` -> `Account`

kclowes (commented):
I haven't totally looked through, but wanted to group these comments together, so maybe they make some sense without other noise :)

Edit: that didn't quite work as I expected, ha. [This comment](https://github.com/ethereum/eth-account/pull/283#discussion_r1669331735) belongs with this group.

eth_account/account_local_actions.py
kclowes: I was playing around with this a little, and I think it makes more sense just to keep Account separate, i.e. not inheriting from `AccountLocalActions`. I'm not sure what we gain from the `AccountLocalActions` abstract class. If we remove it, then `Account` can be its own separate entity and we don't need `AccountLocalActions` any more. What do you think about that? I may be missing something too.

eth_account/types.py
```diff
@@ -0,0 +1,31 @@
...
+
+# from web3.types
+class AccessListEntry(TypedDict):
+    address: HexStr
+    storageKeys: Sequence[HexStr]
+
+
+AccessList = Sequence[AccessListEntry]
+
+TransactionDictType = Dict[str, Union[AccessList, bytes, HexStr, int]]
```
kclowes, line 23: Do you think it's worth spelling out what all the keys could be here? `nonce`, `to`, `from`, etc.

eth_account/account_local_actions.py
pacrob: It's the circular import - `LocalAccount` needs an `Account` as for its `_public_api`, but an Account needs to be able to return instances of `LocalAccount`

pacrob: The other idea was to split Account into 2 classes - signing methods and creation methods, with the full method implementation in each. Those could be joined into what is currently `Account`, and `LocalAccount` would just take the signing one.

kclowes: > It's the circular import - LocalAccount needs an Account as for its _public_api, but an Account needs to be able to return instances of LocalAccount

Ugh, just saw that. Sorry for the goose chase 🪿 . I think this is okay for now. Will continue reviewing!

eth_account/_utils/signing.py
```diff
@@ -103,48 +121,55 @@ def extract_chain_id(raw_v):
...
         v = v_raw + V_OFFSET
     else:
         v = v_raw + CHAIN_ID_OFFSET + 2 * chain_id
     return v
 
 
-def sign_transaction_hash(account, transaction_hash, chain_id):
+def sign_transaction_hash(
+    account: PrivateKey, transaction_hash: Bytes32, chain_id: Optional[int] = None
+) -> Tuple[int, int, int]:
```
kclowes, line 148: I think this can be: 
```suggestion
) -> Tuple[int, Bytes32, Bytes32]:
```
but feel free to ignore if not :)

eth_account/_utils/signing.py
```diff
@@ -103,48 +121,55 @@ def extract_chain_id(raw_v):
...
+            to_bytes,
+            _pad_to_eth_word,
+        ),
     )
 
 
-def sign_message_hash(key, msg_hash):
+def sign_message_hash(
+    key: PrivateKey, msg_hash: Bytes32
+) -> Tuple[int, int, int, bytes]:
```
kclowes, line 172: I think `r` and `s` here can be `Bytes32`

kclowes (approved):
LGTM! I left a couple nits around making `r` and `s` values from `int` -> `Bytes32`, but feel free to take or leave if that doesn't make sense!

eth_account/types.py
fselmo: Agreed...I wonder if it's worth moving the *web3.py* `TxParams` to *eth_typing* and using it here as well since we're getting ready for a beta round before `v7` there?

fselmo: Don't mind me, I think these are different expected types. But perhaps the same idea with a `TypedDict` with `total=False` here would be more reader friendly.

eth_account/types.py
```diff
@@ -0,0 +1,31 @@
...
+from hexbytes import (
+    HexBytes,
+)
+
+Blobs = Sequence[Union[bytes, HexBytes]]
+Bytes32 = bytes
+PrivateKeyType = Union[Bytes32, int, HexStr, PrivateKey]
+
+
+# from web3.types
```
fselmo: nit: While we're in a beta round, maybe this is worth putting into the current *eth-typing* beta and using it in both *web3.py* and here?

I'd encourage any other types that are borrowed to be used in that manner as well. That would also remove any sort of weirdness coming out of the same type name possibly coming from two different sources within *web3.py* if that ever comes up.

eth_account/_utils/signing.py
pacrob: I put a breakpoint in there and the values are `int`s.

fselmo (approved):
lgtm, as we mentioned let's try to address some of the type narrowing and sharing, by the way of refactoring to *eth-typing*, in a subsequent PR before a release.

PR #285, opened 2024-06-26
Set remaining mypy setting to true

### LINKED ISSUE #101: Enable stricter mypy checks

pacrob:
### What was wrong?

Tighten up typing!

Closes #101

### How was it fixed?

Set `mypy` rules `disallow_any_generics`, `disallow_subclassing_any` and `disallow_untyped_decorators` to `true`, fix errors

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/_utils/transaction_utils.py
+++ b/eth_account/_utils/transaction_utils.py
@@ -2,6 +2,7 @@ from typing import (
     Any,
     Dict,
     Sequence,
+    Tuple,
 )
 
 from toolz import (
@@ -74,7 +75,9 @@ def transaction_rpc_to_rlp_structure(dictionary: Dict[str, Any]) -> Dict[str, An
     return dictionary
 
 
-def _access_list_rpc_to_rlp_structure(access_list: Sequence) -> Sequence:
+def _access_list_rpc_to_rlp_structure(
+    access_list: Sequence[Dict[str, Any]]
+) -> Tuple[Tuple[str, Tuple[str]], ...]:
     if not is_rpc_structured_access_list(access_list):
         raise ValueError(
             "provided object not formatted as JSON-RPC-structured access list"
@@ -104,7 +107,9 @@ def transaction_rlp_to_rpc_structure(dictionary: Dict[str, Any]) -> Dict[str, An
     return dictionary
 
 
-def _access_list_rlp_to_rpc_structure(access_list: Sequence) -> Sequence:
+def _access_list_rlp_to_rpc_structure(
+    access_list: Sequence[Any],
+) -> Sequence[Dict[str, Any]]:
     if not is_rlp_structured_access_list(access_list):
         raise ValueError("provided object not formatted as rlp-structured access list")
 
--- a/eth_account/_utils/transaction_utils.py
+++ b/eth_account/_utils/transaction_utils.py
@@ -1,8 +1,6 @@
 from typing import (
     Any,
     Dict,
-    Sequence,
-    Tuple,
 )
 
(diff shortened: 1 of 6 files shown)


eth_account/_utils/transaction_utils.py
```diff
@@ -104,7 +107,9 @@ def transaction_rlp_to_rpc_structure(dictionary: Dict[str, Any]) -> Dict[str, An
     return dictionary
 
 
-def _access_list_rlp_to_rpc_structure(access_list: Sequence) -> Sequence:
+def _access_list_rlp_to_rpc_structure(
+    access_list: Sequence[Any],
```
kclowes: It looks like this is a pretty specific structure based on `t[0]` and `t[1]` from line 117. Looks like a two-tuple with address as the first and storage keys as the second. Maybe something like:
```suggestion
    access_list: Tuple[Address, Sequence[Bytes32]]
```

eth_account/_utils/transaction_utils.py
```diff
@@ -104,7 +107,9 @@ def transaction_rlp_to_rpc_structure(dictionary: Dict[str, Any]) -> Dict[str, An
...
+) -> Sequence[Dict[str, Any]]:
```
kclowes: I think this can just be `AccessList`

kclowes (commented):
I think we can use the shiny new `AccessList` here. Left a couple inline comments, but looks good otherwise!

eth_account/_utils/transaction_utils.py
pacrob: We're mostly using `HexStr` for addresses here. I'll be taking another look at all this to see if the `TxData` typed dict from web3 could be shared with eth-account too.

kclowes (approved):
LGTM! 💥

eth_account/types.py
```diff
@@ -19,5 +20,6 @@
 PrivateKeyType = Union[Bytes32, int, HexStr, PrivateKey]
 
 AccessList = Sequence[Dict[str, Union[HexStr, Sequence[HexStr]]]]
+RLPStructuredAccessList = Tuple[Tuple[HexStr, Tuple[HexStr, ...]], ...]
```
fselmo: I wonder if these should be `Sequence` as well? Rather than `Tuple`

pacrob: updated

eth_account/types.py
```diff
@@ -19,5 +20,6 @@
 PrivateKeyType = Union[Bytes32, int, HexStr, PrivateKey]
 
 AccessList = Sequence[Dict[str, Union[HexStr, Sequence[HexStr]]]]
+RLPStructuredAccessList = Sequence[Tuple[HexStr, Tuple[HexStr, ...]]]
```
fselmo: Let's use `Sequence` here as well

```suggestion
RLPStructuredAccessList = Sequence[Sequence[HexStr, Sequence[HexStr, ...]]]
```

pacrob: updated

fselmo (approved):
lgtm!

PR #288, opened 2024-07-09
Typed dict for transaction dicts

pacrob:
### What was wrong?

Transaction Dicts could use better typing, especially to play better with web3.py

### How was it fixed?

This is a test PR (at least to start), copying the `TxParams` typed dict and any needed sub-types over from web3. If it makes sense, next step will be to pull those types into eth-typing to be shared across both eth-account and web3.

### Todo:

- [ ] Clean up commit history
- [ ] Add or update documentation related to these changes
- [ ] Add entry to the [release notes](https://github.com/ethereum/eth-account/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/_utils/legacy_transactions.py
+++ b/eth_account/_utils/legacy_transactions.py
@@ -35,7 +35,7 @@ from eth_account.typed_transactions import (
 )
 from eth_account.types import (
     Blobs,
-    TransactionDictType,
+    TxParams,
 )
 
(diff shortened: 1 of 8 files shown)


pacrob:
This will be a significant change and need to be coordinated with possible changes to `web3` - moving types to be used by both to `eth-typing` or similar. Saving for a `web3 v8` activity.

PR #302, opened 2024-09-19
Don't trigger HDAccount DeprecationWarning if Language enum is passed in

kclowes:
### What was wrong?
We made the language that gets passed into the HD Accounts an `Enum`, and deprecated passing the string representation in. However, the DeprecationWarning was getting raised even though an Enum got passed.

### How was it fixed?
Removed extraneous DeprecationWarnings, and changed the return value of `detect_language` because I didn't see it documented anywhere, and therefore I am assuming it was meant to be a private method. Also added a new `list_languages_enum` method. Could stand to come up with a better name for that method, if you have any ideas. Additionally, added a few tests.
### Todo:

- [x] Clean up commit history
- [x] Add entry to the [release notes](https://github.com/ethereum/eth-account/blob/main/newsfragments/README.md) - only added a feature newsfragment because the bugfix portion hadn't been released.

#### Cute Animal Picture

### DIFF
--- a/eth_account/types.py
+++ b/eth_account/types.py
@@ -35,3 +35,8 @@ class Language(enum.Enum):
     JAPANESE = "japanese"
     KOREAN = "korean"
     SPANISH = "spanish"
+
+    def __lt__(self, other):
+        if self.__class__ is other.__class__:
+            return self.value < other.value
+        return NotImplemented
--- a/eth_account/types.py
+++ b/eth_account/types.py
@@ -1,5 +1,6 @@
 import enum
 from typing import (
+    Any,
     Dict,
     Sequence,
     Union,
@@ -36,7 +37,7 @@ class Language(enum.Enum):
     KOREAN = "korean"
     SPANISH = "spanish"
 
-    def __lt__(self, other):
-        if self.__class__ is other.__class__:
+    def __lt__(self, other: Any) -> Any:
+        if other.__class__ is Language:
             return self.value < other.value
         return NotImplemented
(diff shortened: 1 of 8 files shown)


reedsa (approved):
lgtm

eth_account/types.py
```diff
@@ -35,3 +36,8 @@ class Language(enum.Enum):
     JAPANESE = "japanese"
     KOREAN = "korean"
     SPANISH = "spanish"
+
+    def __lt__(self, other: Any) -> Any:
```
pacrob, line 40: Why is this added?

pacrob (approved):
Just one question, otherwise lgtm! Just to note, I'm not able to reproduce the `DeprecationWarning` using the `enum` on current `main`. How did you get it to show?

fselmo (approved):
👍🏼

eth_account/types.py
kclowes: In [this line](https://github.com/ethereum/eth-account/pull/302/files#diff-721e355e9fe8b99722fbb9bfbe182fb03c0929f4601b48ab92ba064d8d14a47cR150), we are sorting the result alphabetically, and the enum didn't know how to do that without adding the `__lt__` method.

kclowes:
> Just to note, I'm not able to reproduce the DeprecationWarning using the enum on current main. How did you get it to show?

In your terminal: 
`$ export PYTHONWARNINGS=always`

and you should see it

pacrob:
> and you should see it

Still doesn't show for me, but no reason to hold this up. It's still an improvement over existing!

PR #306, opened 2024-11-25
add sign_typed_data to local signer

### LINKED ISSUE #304: Allow calling `sign_typed_data` from a `LocalAccount`
### What happened?

While instantiated `LocalAccount`s can sign messages and transactions, they cannot yet sign typed data. This is [only available](https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.account.Account.sign_typed_data) through the base `Account` class, having to pass in the key.

It would be helpful to call this through the context of an existing account, so that one could simply do:
```py
my_account.sign_typed_data(full_message=payload)
```

_No response_

### Full error output

```shell
AttributeError: 'LocalAccount' object has no attribute 'signed_typed_data'
```

### Fill this section in if you know how this could or should be fixed

_No response_

### eth-account Version

0.13.4

### Python Version

_No response_

### Operating System

_No response_

### Output from `pip freeze`

_No response_

pacrob:
### What was wrong?

`sign_typed_data` was available in `Account` but not `LocalAccount`.

Closes #304 

### How was it fixed?

Added it!

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/account_local_actions.py
+++ b/eth_account/account_local_actions.py
@@ -72,3 +72,15 @@ class AccountLocalActions(ABC):
         blobs: Optional[Blobs] = None,
     ) -> SignedTransaction:
         pass
+
+    @combomethod
+    @abstractmethod
+    def sign_typed_data(
+        self,
+        private_key: PrivateKeyType,
+        domain_data: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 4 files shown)


fselmo (approved):
lgtm!

PR #310, opened 2025-01-29
bump keyfile and abi versions, fix new lint errors

### LINKED ISSUE #305: Pin `eth-keyfile` dependency to >=0.9.0

The next release of `eth-keyfile` will be `0.9.0` containing breaking changes. Once it's released, change that to the bottom pin here in `eth-account`.

pacrob:
### What was wrong?

`eth-keyfile v0.9.0` was released, so we can now bump its dep here to `>=0.9.0`.

### How was it fixed?

Bumped and fixed minor lint errors. Removed the beta from the `eth-abi` dep too.
Closes #305 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -200,7 +200,7 @@ class Account(AccountLocalActions):
         # type ignored because eth_keyfile appears to be using the wrong type for
         # the password arg.
         # once fixed there, this should error and can be removed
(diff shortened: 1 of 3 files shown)


kclowes (commented):
I could be persuaded otherwise, but I think we should keep this stacked until we're ready to drop py3.8 and jump into a breaking changes cycle across the libraries. It's a small change, but I'm concerned we'll create a dependency hell for users with the eth-keyfile bump. But it looks good otherwise!

PR #316, opened 2025-02-21
implement set_code_transaction for EIP-7702

### LINKED ISSUE #315: A new type of transaction is needed for EIP-7702

EIP-7702 style stransactions will be available on the public testnet on feb 28th 2025
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md

Most of the changes needed for EIP-7702 are boils down to having a new type of transaction able to carry an authorizationList

joseluu:
### What was wrong?

Closes #315 

### How was it fixed?
a new type of transaction is implemented named set_code_transaction as well as the required validators and serializers
2 tests added
tested on pectra-devnet-6 https://pectra-devnet-6.ethpandaops.io/

#### Cute Animal Picture

satoshi cat


eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+        ("nonce", big_endian_int),
+        ("yParity", big_endian_int),
+        ("signerR", big_endian_int),
+        ("signerS", big_endian_int),
+    )
+
+
+class SetCodeTransaction(_TypedTransactionImplementation):
+    """
+    Represents a dynamic fee transaction access per EIP-1559.
```
fselmo: This should describe the new transaction per EIP-7702

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+    """
+
+    # This is the second transaction to implement the EIP-2718 typed transaction.
```
fselmo:
```suggestion
    # This is the third transaction to implement the EIP-2718 typed transaction.
```

eth_account/typed_transactions/typed_transaction.py
```diff
@@ -108,8 +114,11 @@ def from_bytes(cls, encoded_transaction: HexBytes) -> "TypedTransaction":
...
         transaction: Union[
-            "DynamicFeeTransaction", "AccessListTransaction", "BlobTransaction"
-        ]
+            "DynamicFeeTransaction",
+            "AccessListTransaction",
+            "BlobTransaction",
+            "SetCodeTransaction",
+         ]
 
         if encoded_transaction[0] == AccessListTransaction.transaction_type:
```
fselmo: Unrelated but we should probably extract `encoded_transaction[0]` as `tx_type` here by now. We re-use it enough and it makes the code more readable.

```suggestion
        tx_type = encoded_transaction[0]

        if tx_type == AccessListTransaction.transaction_type:
        ... # use tx_type for all cases here
```

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+        return cast(bytes, hash)
+
+    def payload(self) -> bytes:
+        """
+        Returns this transaction's payload as bytes.
+
+        Here, the transaction payload is:
+
+            TransactionPayload = rlp([chainId,
```
fselmo, line 270: This should be updated to match the actual payload for set code tx.

eth_account/typed_transactions/base.py
```diff
@@ -74,6 +74,23 @@
         "maxFeePerGas": hexstr_if_str(to_int),
         "maxFeePerBlobGas": hexstr_if_str(to_int),
         "blobVersionedHashes": apply_formatter_to_array(hexstr_if_str(to_bytes)),
+        "authorisationList": apply_formatter_to_array(
```
kclowes: What do you think about making this a tuple instead of a dict? Then we don't have to do as much gymnastics with the rlp -> rpc conversion.

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+            pipe(
+                self,
+                rlp.encode,
+                lambda val: bytes([self.__class__.magic]) + val,  # (0x05 || rlp([...]))
+                keccak,
+            )
+        )
+
+
+class SignedAutorizationRLP(HashableRLP):
```
kclowes:
```suggestion
class SignedAuthorizationRLP(HashableRLP):
```

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+    @classmethod
+    def from_dict(
+        cls, dictionary: Dict[str, Any], blobs: Optional[Blobs] = None
+    ) -> "SetCodeTransaction":
+        """
+        Builds a SetCodeTransaction from a dictionary.
+        Verifies that the dictionary is well formed.
+        """
+        if blobs is not None:
+            raise ValueError("Blob data is not supported for `DynamicFeeTransaction`.")
```
kclowes:
```suggestion
            raise ValueError("Blob data is not supported for `SetCodeTransaction`.")
```

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+
```
kclowes, line 191: I think we want to add the line: 
`dictionary = set_transaction_type_if_needed(dictionary)` somewhere in here

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+    def from_bytes(cls, encoded_transaction: HexBytes) -> "SetCodeTransaction":
+        """Builds an SetCodeTransaction from a signed encoded transaction."""
+        if not isinstance(encoded_transaction, HexBytes):
+            raise TypeError(f"expected Hexbytes, got type: {type(encoded_transaction)}")
+        if not (
+            len(encoded_transaction) > 0
+            and encoded_transaction[0] == cls.transaction_type
+        ):
+            raise ValueError("unexpected input")
+        # Format is (0x01 || TransactionPayload)
```
kclowes:
```suggestion
        # Format is (0x04 || TransactionPayload)
```

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,297 @@
...
+        dictionary = self.dictionary.copy()
+        dictionary["type"] = self.__class__.transaction_type
+        return dictionary
+
+    def hash(self) -> bytes:
+        """
+        Hashes this AccessListTransaction to prepare it for signing.
+        As per the EIP-2930 specifications, the signature is a secp256k1 signature over
+        ``keccak256(0x01 || rlp([chainId, nonce, gasPrice, gasLimit,
+        to, value, data, accessList])).``
```
kclowes: needs to be updated to `SetCode`

newsfragments/315.feature.rst
```diff
@@ -0,0 +1,47 @@
+set_code_transaction for EIP-7702
+
+### Overview
```
kclowes: Thank you for adding such detailed documentation! This type of documentation belongs in a file other than the newsfragment file. If it stays here, it will just end up in the release notes. You'll need to follow the pattern in docs/eth_account.typed_transactions.rst to add the methods and comments from the code. You may be able to copy/paste this into that same doc file too, otherwise maybe it should be it's own page.

eth_account/typed_transactions/base.py
```diff
@@ -74,6 +74,23 @@
...
+                    "chainId": hexstr_if_str(to_int),
+                    "nonce": hexstr_if_str(to_int),
+                    "address": apply_one_of_formatters(
+                        (
+                            (is_string, hexstr_if_str(to_bytes)),
+                            (is_bytes, identity),
+                        )
+                    ),
+                    "yParity": hexstr_if_str(to_int),
+                    "signerR": hexstr_if_str(to_int),
```
kclowes: I have a slight preference to simply call this `r`, and `s`, but could be convinced otherwise.

eth_account/_utils/transaction_utils.py
```diff
@@ -68,10 +75,21 @@ def transaction_rpc_to_rlp_structure(dictionary: Dict[str, Any]) -> Dict[str, An
     Convert a JSON-RPC-structured transaction to an rlp-structured transaction.
     """
     access_list = dictionary.get("accessList")
+    authorisation_list = dictionary.get("authorisationList")
     if access_list:
         dictionary = dissoc(dictionary, "accessList")
         rlp_structured_access_list = _access_list_rpc_to_rlp_structure(access_list)
         dictionary = assoc(dictionary, "accessList", rlp_structured_access_list)
+        access_list = dictionary.get("accessList")
```
kclowes: I don't see this getting used anywhere.

tests/core/test_typed_transactions.py
```diff
@@ -353,6 +422,8 @@
     "blob-no-access-list",
     "blob-int-values-and-access-list",
     "blob-no-explicit-type",
+    "sc-standard-case",
+    "sc-and-al",
 ]
```
kclowes, line 509: Nice tests!
- It would be good if we could find at least one test with an expected hash from a trusted source like the tests repo or something similar. I'll see if I can find something here too.
- I'd like to see one more test case with more than one authorizationList too.

kclowes (commented):
Thanks @joseluu! I had a very similar branch going, but happy to take yours :) It looks like you're using both `Authorization` and `Authorisation`. If it's all right with you, I would prefer to standardize on `Authorization`, because that's how it's written in the EIP.

joseluu:
Thanks for your review, I will fix all this

eth_account/typed_transactions/base.py
fselmo: I have a stronger preference for `r` and `s`. Sticks to the EIP and the library. I'd prefer not to deviate there.

fselmo (commented):
Had a quick initial pass here. I had similar comments to @kclowes. We should search + replace for `Authorisation` -> `Authorization` and `authorisation` -> `authorization`. Left some other comments inline but thanks for getting this going!

eth_account/typed_transactions/base.py
joseluu: the signed authorization is going to be circulated among the participants, so as to avoid confusion with the initial users, I made it this way. it is a matter of taste, you tell me.

fselmo: I would prefer to keep `r` and `s` here

eth_account/typed_transactions/base.py
joseluu: accessList being array it may appear inconsistent

fselmo:
Thanks so much for this @joseluu! I added the functionality to the `LocalAccount` which is nice if you want to use `acct.sign_authorization()` without passing in a private key (and this should bubble up to web3.py too which will be quite nice).

I also added some more tests and I ended up adding the model as a pydantic model because I think we can really leverage serialization better across the stack to web3.py. This is an attempt to start down that line of thinking.

I also added the `authorization` (address that authorized) as well as the signature and hash to the returned signed object. I think this can be handy to include if we already have that data available while signing.

I'm going to fix the docs issues next and then I am good with merging this! Thanks again. It's always nice to get some solid contributions like these 😃

eth_account/account.py
```diff
@@ -1043,3 +1054,99 @@ def sign_typed_data(
...
+        :param dict authorization_dict: the required keys are: chainId, address, nonce
+        :param private_key: the private key to sign the data with
+        :type private_key: hex str, bytes, int or :class:`eth_keys.datatypes.PrivateKey`
+        :returns: the dictionary with the signature fields added, suitable for inclusion in a EIP-7702 transaction
+        :rtype: dict
+
+        .. NOTE::
+            You need to get signed one or more signed authorizations from an EOA willing to have a smart contract code associated with the EOA, this the essence of EIP-7702
+
+            in the code below in the variable authorization_to_sign:
```
pacrob:
```suggestion
            in the code below in the variable `auth`:
```

eth_account/account.py
```diff
@@ -1043,3 +1054,99 @@ def sign_typed_data(
...
+            - chainId is the chain id of the chain where the EOA is located or 0 if the authorization is for all chains
```
pacrob:
```suggestion
            - `chainId` is the chain id of the chain where the EOA is located or 0 if the authorization is for all chains
```

eth_account/account.py
```diff
@@ -1043,3 +1054,99 @@ def sign_typed_data(
...
+            - address is the address of the smart contract code to be associated with the EOA, the address format is bytes
```
pacrob:
```suggestion
            - `address` is the address of the smart contract code to be associated with the EOA, the address format is bytes
```

eth_account/account.py
```diff
@@ -1043,3 +1054,99 @@ def sign_typed_data(
...
+            - nonce is the nonce of the EOA, it is used to prevent replay attacks
```
pacrob:
```suggestion
            - `nonce` is the nonce of the EOA, it is used to prevent replay attacks
```

eth_account/account.py
kclowes: You need a new line here to get the bullet list to render correctly:

```suggestion
            in the code below in the variable authorization_to_sign:
            
```

kclowes: this is `rst`, so I think you'll want \`\`auth``. Same with the below

eth_account/account.py
```diff
@@ -1043,3 +1054,99 @@ def sign_typed_data(
...
+            >>> tx = {
+            ...     "gas": 100000,
+            ...     "maxFeePerGas": 2000000000,
+            ...     "maxPriorityFeePerGas": 2000000000,
+            ...     "data": "0x616263646566",
+            ...     "nonce": 34,
+            ...     "to": "0x09616C3d61b3331fc4109a9E41a8BDB7d9776609",
+            ...     "value": "0x5af3107a4000",
+            ...     "accessList": (),
+            ...     "authorizationList": [auth_signed.as_rpc_object()],
```
kclowes: This feels clunky. Would it be a huge pain to handle this on the other side so the user can just pass `auth_signed`?

tests/core/test_mnemonic.py
```diff
@@ -136,7 +143,7 @@ def test_generation_with_string(lang, num_words):
     # NOTE: Sometimes traditional chinese can return characters that are also
     # valid simplified chinese characters. In that scenario, the detection
     # algorithm will assume simplified.
-    if lang == "chinese_simplified":
+    if lang == "chinese_traditional":
```
kclowes: I'm curious to know the reasoning behind changing this?

eth_account/typed_transactions/set_code_transaction.py
kclowes: I think this still holds?

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,282 @@
...
+
+    transaction_field_defaults = {
+        "type": b"0x4",
+        "chainId": 0,
+        "nonce": 0,
+        "to": b"",
+        "value": 0,
+        "data": b"",
+        "accessList": [],
+        "authorizationList": [],
```
kclowes, line 130: It looks like sometimes we allow a List and sometimes we allow a Tuple for authorizationList, and we should make sure that we're consciously allowing both. I don't have a strong preference for either allowing both or picking one, but just wanted to flag it.

eth_account/signers/local.py
```diff
@@ -158,3 +158,9 @@ def sign_typed_data(
                 full_message=full_message,
             ),
         )
+
+    def sign_authorization(self, authorization: Dict[str, Any]) -> SignedMessage:
```
kclowes, line 162: Is there a use case where we'd want to pass an authorization list here? For example either a new method something like `sign_authorization_list` or `sign_authorizations(self, authorizations: List[Dict[str, Any]]) ...`

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,282 @@
...
+        """Returns this transaction as a dictionary."""
+        dictionary = self.dictionary.copy()
+        dictionary["type"] = self.__class__.transaction_type
+        return dictionary
+
+    def hash(self) -> bytes:
+        """
+        Hashes this SetCodeTransaction to prepare it for signing.
+        As per the EIP-7702 specifications, the signature is a secp256k1 signature over
+        ``keccak256(0x04 || rlp([chainId, nonce, gasPrice, gasLimit, to, value, data,
```
kclowes:
```suggestion
        ``keccak256(0x04 || rlp([chainId, nonce, max_priority_fee_per_gas, max_fee_per_gas, gasLimit, to, value, data,
```

kclowes (approved):
I left a few comments, but nothing blocking for when you're ready to merge @fselmo!

eth_account/account.py
fselmo: I currently have [an issue open](https://github.com/ethereum/web3.py/issues/3650) to be able to deprecate this and handle it implicitly in `web3.py`. I'm not really sure how else it can be "not clunky". Open to ideas. I'd rather have the information that is already there in this object and handle the serialization as JSON-RPC. 

It's not even as simple as removing the authorization address, the signature, and the hash from this object since JSON-RPC uses camelCase. We'd have to completely conform to `camelCase` within the Python class, which we moved away from for `SignedTransaction` (`rawTransaction` -> `raw_transaction`).

eth_account/signers/local.py
fselmo: Authorizations can be signed by different accounts. I think that would abstract a fairly small step.

```py
authorization_list = [
    acct.sign_authorization(auth).as_rpc_object() for auth in (
        {"address": ..., "nonce": ..., "chainId": ...},
        {"address": ..., "nonce": ..., "chainId": ...},
    )
]
```

I lean towards keeping it focused on single auths tbh. Maybe we can see if there's demand first.

eth_account/typed_transactions/set_code_transaction.py
fselmo: I personally feel like the original API for these transactions has friction points. But it seems the way that it's meant to be used is to use the `TypedTransaction.from_dict()` which [already has this step](https://github.com/ethereum/eth-account/blob/main/eth_account/typed_transactions/typed_transaction.py#L80-L106). So this feels redundant. Though we do have certain TypedTransaction implementation classes that also [add the type](https://github.com/ethereum/eth-account/blob/main/eth_account/typed_transactions/blob_transactions/blob_transaction.py#L212-L213), and [some that don't](https://github.com/ethereum/eth-account/blob/main/eth_account/typed_transactions/dynamic_fee_transaction.py#L138-L168).

I do lean toward adding it to reduce this friction as much as possible, and I'll add it in this PR, just wanted to point out where some inconsistencies lie.

eth_account/account.py
fselmo: Come to think of it, I think this probably needs to be addressed in the `from_dict()` of the `SetCodeTransaction`. I'm going to see how to best structure this but it may very well be that to sign the transaction in the first place, it always passes through `from_dict`.

tests/core/test_mnemonic.py
joseluu: If actual langage (variable lang)  is "chinese_traditional"  and if the previous code were active, the line 149 may fail
149:     assert Mnemonic.detect_language(mnemonic) == Language(lang)

This is because, as noted, chinese_traditional contains simplified characters, it may therefore happen that the detect_langage function returns "chinese_simplified" when it is not the case, leading to assert failure at line 139
This case happened to me while executing the tests

eth_account/account.py
fselmo: Clunky it was indeed. Turns out my last comment was correct. We never really need to get all the way over to `web3.py` for this conversion, it should be happening inside `from_dict()`. I updated it here and added tests to make sure that the Pydantic classes are properly serialized as JSON-RPC objects.

fselmo:
@kclowes this should be updated. Requesting a re-review for when you have some time.

eth_account/account.py
```diff
@@ -675,11 +686,11 @@ def sign_transaction(
         Sign a transaction using a local private key.
 
         It produces signature details and the hex-encoded transaction suitable for
-        broadcast using :meth:`w3.eth.sendRawTransaction()
-        <web3.eth.Eth.sendRawTransaction>`.
+        broadcast using :meth:`w3.eth.send_raw_transactions()
```
kclowes: should be:
```suggestion
        broadcast using :meth:`w3.eth.send_raw_transaction()
```

eth_account/account.py
```diff
@@ -675,11 +686,11 @@ def sign_transaction(
...
+        <web3.eth.Eth.send_raw_transactions>`.
```
kclowes: same here:
```suggestion   <https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.send_raw_transaction>.
```

eth_account/account.py
```diff
@@ -675,11 +686,11 @@ def sign_transaction(
...
 
         To create the transaction dict that calls a contract, use contract object:
-        `my_contract.functions.my_function().buildTransaction()
+        `my_contract.functions.my_function().build_transaction()
         <http://web3py.readthedocs.io/en/latest/contracts.html#methods>`_
```
kclowes: This link should be: 
https://web3py.readthedocs.io/en/stable/web3.contract.html#methods

eth_account/signers/local.py
kclowes: That sounds like a good plan to me! 👍

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,289 @@
...
+        dictionary = self.dictionary.copy()
+        dictionary["type"] = self.__class__.transaction_type
+        return dictionary
+
+    def hash(self) -> bytes:
+        """
+        Hashes this SetCodeTransaction to prepare it for signing.
+        As per the EIP-7702 specifications, the signature is a secp256k1 signature over
+        ``keccak256(0x04 || rlp([chainId, nonce, max_priority_fee_per_gas,
+        max_fee_per_gas, gasLimit, gasLimit, to, value, data, accessList,
```
kclowes:
```suggestion
        max_fee_per_gas, gasLimit, to, value, data, accessList,
```

eth_account/typed_transactions/set_code_transaction.py
```diff
@@ -0,0 +1,289 @@
...
+        Returns this transaction's payload as bytes.
+
+        Here, the transaction payload is:
+
+            TransactionPayload = rlp([chainId,
+            nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit,
+            to, value, data,
+            accessList,
+            authorizationsList,
+            signatureYParity, signatureR, signatureS])
```
kclowes, line 275: Did we decide this should be `y_parity, r, s` or am I thinking of something else?

tests/core/test_mnemonic.py
kclowes: Thank you for the clarification. That makes sense to me!

tests/core/test_mnemonic.py
```diff
@@ -136,7 +143,7 @@ def test_generation_with_string(lang, num_words):
     # NOTE: Sometimes traditional chinese can return characters that are also
     # valid simplified chinese characters. In that scenario, the detection
     # algorithm will assume simplified.
```
kclowes:
```suggestion
    # algorithm will assume traditional.
```

kclowes (approved):
Code LGTM, mostly left doc nits!

eth_account/account.py
fselmo: changed to a `:meth:` reference and tested it. I think that may be better so broken links don't need to be updated 👍🏼

eth_account/typed_transactions/set_code_transaction.py
fselmo: The payload definition in the EIP makes this distinction. I think this is fine. I'd just rather not have the actual fields named `signature...` or `signer...` and have them instead be `r`, `s`, `y_parity` / `yParity`, as both the EIP and our code currently do.

tests/core/test_mnemonic.py
fselmo: I also ended up fixing the enum case. I ran 1000 tests of traditional and simplified and I was still getting some errors. I ended up changing the test to more explicitly test that if there is a detect_language failure it is only for this one specific case and it's more clear for a reader of the code now.

I tested this against 1000 of each many times over and this seems to be working well.

PR #318, opened 2025-03-28
Some refactoring for pydantic models:

fselmo:
### What was wrong?

- Omit problematic fields from the json schema. This was causing
  some issues with fast api for some users that are serving the models
  over REST.

- Better organize and simplify the serialization of these models. Models
  should be indepdendently serializable and deserializable, as well
  as when creating a ``TypedTransaction.from_dict()``, NOT just when
  creating the ``TypedTransaction`` object. There was an issue where
  an ``@field_serializer`` was not being applied correctly because it was
  below the ``@classmethod``. It needs to be above the ``@classmethod``.
  This led to some unnecessary gymnastics to make the models work (see:
  ``recursive_model_dump()``). This method should probably be deprecated
  eventually.

- Add tests for serialization of the new datastructures / pydantic models
  and for generating the json schema.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`0_o`

### DIFF
--- a/eth_account/datastructures.py
+++ b/eth_account/datastructures.py
@@ -1,12 +1,14 @@
 from typing import (
     Any,
+    Dict,
     NamedTuple,
-    Set,
     SupportsIndex,
     Tuple,
+    Type,
     Union,
     overload,
 )
+import warnings
 
 from eth_keys.datatypes import (
     Signature,
@@ -23,10 +25,21 @@ from hexbytes import (
 from pydantic import (
     BaseModel,
     ConfigDict,
+    Field,
     field_serializer,
 )
 from pydantic.alias_generators import (
     to_camel,
+)
+from pydantic.json_schema import (
+    DEFAULT_REF_TEMPLATE,
+    GenerateJsonSchema,
+    JsonSchemaMode,
+    JsonSchemaValue,
+)
+from pydantic_core import (
+    CoreSchema,
+    PydanticOmit,
 )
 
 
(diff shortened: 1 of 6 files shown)


eth_account/datastructures.py
```diff
@@ -153,13 +171,11 @@ class SignedSetCodeAuthorization(CustomPydanticModel):
...
     s: int
-    signature: Signature
-    authorization_hash: HexBytes
-
-    _exclude = {"signature", "authorization_hash", "authority"}
+    signature: Signature = Field(exclude=True)
+    authorization_hash: HexBytes = Field(exclude=True)
 
-    @classmethod
     @field_serializer("address")
```
fselmo, line 177: keep this one in mind... this was the cause of unnecessary pain :)

fselmo:
cc: @Royalnine 👀

royalnine (commented):
thank you @fselmo 🙏

tests/core/test_datastructures.py
```diff
@@ -0,0 +1,70 @@
...
+    PydanticTestClass,
+)
+
+
+@pytest.mark.parametrize(
+    "pydantic_model,expected",
+    (
+        (
+            TEST_SIGNED_AUTHORIZATION,
+            {
```
reedsa: Nit: Expected data could be pulled into _test_utils

reedsa (approved):
No concerns, minor nit but up to you if it's worth using a constant.

tests/core/test_datastructures.py
fselmo: def worth it as it's already used in two places 👍🏼 ... updated

PR #319, opened 2025-04-03
clarify auth signing docs; nonce + 1 & data

fselmo:
### What was wrong?

The docs were just copy pasted from another example (sorry), so they really don't serve as a great reference. We should clarify that the nonce in an auth for a tx that will be sent by the same signer should have a + 1... and we should clarify what the data should be.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`o_0`

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -1086,28 +1086,30 @@ class Account(AccountLocalActions):
             >>> auth = {
             ...     "chainId": 1337,
             ...     "address": "0x5ce9454909639d2d17a3f753ce7d93fa0b9ab12e",
-            ...     "nonce": 0,
+            ...     # If the tx sender is the same as the authority, the nonce needs to be 1 higher than the current (transaction) nonce.
+            ...     # This is because the nonce is increased before tx execution and again when processing the authorization.
+            ...     "nonce": 1,
             ... }
 
             >>> signed_auth = Account.sign_authorization(auth, key)
             >>> signed_auth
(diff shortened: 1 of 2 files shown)


eth_account/account.py
```diff
@@ -1086,28 +1086,30 @@ def sign_authorization(
             >>> auth = {
             ...     "chainId": 1337,
             ...     "address": "0x5ce9454909639d2d17a3f753ce7d93fa0b9ab12e",
-            ...     "nonce": 0,
+            ...     # If the tx sender is the same as the authority, the nonce needs to be 1 higher than the current (transaction) nonce.
```
reedsa, line 1089: Curious what thoughts you have about a couple things. I can understand if this content is next to the value here, but wanted to consider what it might be like in the docstring before this example.

Also wondering if it makes sense to plug in `web3.eth.get_transaction_count` instead of a static example value.

fselmo: > what it might be like in the docstring before this example.

This is specific to the example. Not many places make this clear so I feel like in the code it shows exactly where this is meant.

> Also wondering if it makes sense to plug in web3.eth.get_transaction_count instead of a static example value.

I'd like to not rely on web3.py too much here, especially as it is a doctest and I'd like to get this passing without relying on and having to import web3.py for static values such as nonces. I'm all for being more web3.py-friendly on the web3.py side of the documentation.

pacrob (approved):
:sheep:

PR #320, opened 2025-04-14
Inherit from eth-utils CamelModel for pydantic types:

fselmo:
### What was wrong?

- The ``CustomPydanticModel`` used here inspired a more general class to be used across our libraries. That change was recently introduced in ``eth-utils``, so inherit from that class here for now, with a note to use ``CamelModel`` directly for the next major version.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2025-04-14 at 16 15 17

### DIFF
--- a/eth_account/_utils/transaction_utils.py
+++ b/eth_account/_utils/transaction_utils.py
@@ -2,6 +2,9 @@ from typing import (
     Any,
 )
 
+from eth_utils import (
+    CamelModel,
+)
 from toolz import (
     assoc,
     dissoc,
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -55,7 +55,7 @@
         "eth-keyfile>=0.7.0, <0.9.0",
         "eth-keys>=0.4.0",
         "eth-rlp>=2.1.0",
-        "eth-utils>=2.0.0",
+        "eth-utils>=5.3.0",
```
reedsa, line 58: Does this merit a major or minor bump release? Curious what impact this may have.

reedsa (approved):
lgtm, curious to hear thoughts about the next release.

setup.py
kclowes: Yeah, historically we've called a lower pin change like this a major. I think we should wait until our breaking cycle, but could be convinced otherwise :)

kclowes (approved):
🔥 much cleaner, nice! LGTM once tests pass!

PR #326, opened 2025-10-15
Fix BaseModel typing error on Pydantic >= 2.12

kclowes:
### What was wrong?
Pydantic 2.12 was causing mypy to fail. 

### How was it fixed?
Added a `**kwargs` argument to the method definition. Not the most elegant, but it works on both Pydantic 2.11 and 2.12+

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ setup(
         "hexbytes>=1.2.0",
         "rlp>=1.0.0",
         "ckzg>=2.0.0",
-        "pydantic>=2.0.0",
+        "pydantic>=2.0.0,<2.12",  # mypy fails at 2.12.0
     ],
     python_requires=">=3.8, <4",
     extras_require=extras_require,
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ setup(
         "hexbytes>=1.2.0",
         "rlp>=1.0.0",
         "ckzg>=2.0.0",
-        "pydantic>=2.0.0,<2.12",  # mypy fails at 2.12.0
+        "pydantic>=2.0.0",
     ],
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -59,7 +59,7 @@
         "hexbytes>=1.2.0",
         "rlp>=1.0.0",
         "ckzg>=2.0.0",
-        "pydantic>=2.0.0",
+        "pydantic>=2.0.0,<2.12",  # mypy fails at 2.12.0
```
fselmo: Since we only care to replace the `OmitJsonSchema` as the default, I'd also be good with ignoring this with something like:

```py
@classmethod
    def model_json_schema(cls, **kwargs: Any) -> Dict[str, Any]:  # type: ignore[override]  # noqa: E501
        """
        Omits excluded fields from the JSON schema, preventing errors that would
        otherwise be raised by the default schema generator.
        """
        # default to ``OmitJsonSchema`` to prevent errors from excluded fields
        kwargs.setdefault("schema_generator", OmitJsonSchema)
        return super().model_json_schema(**kwargs)
```

fselmo (approved):
Left a suggestion, feel free to take or leave though. lgtm 👍🏼

setup.py
kclowes: I couldn't get the `type: ignore` working on both 2.11 and 2.12, but am able to add a generic `**kwargs: Any` to the params instead of pinning which feels better. lmk if you have objections or think we can tighten it up at all!

fselmo: no this is great 👍🏼

ship it 👌🏼

PR #327, opened 2025-12-01
Add cell proofs for peerdas compat (Fusaka)

fselmo:
### What was wrong?

Related to Issue #
Closes #

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`0_o`


Nomamen (approved):

eth_account/typed_transactions/blob_transactions/blob_transaction.py
```diff
@@ -250,17 +280,25 @@ def from_bytes(cls, encoded_transaction: HexBytes) -> "BlobTransaction":
...
                 transaction_payload
             ).as_dict()
         except rlp.exceptions.ObjectDeserializationError:
-            # If the deserialization fails, we attempt to deserialize as a
-            # `TransactionPayloadBody`, as defined in EIP-4844.
-            dictionary = cls._signed_transaction_serializer.from_bytes(  # type: ignore  # noqa: E501
-                transaction_payload
-            ).as_dict()
+            # Try legacy EIP-4844 format (without wrapper_version, with proofs)
+            try:
```
fselmo, line 287: Really nice, yeah I think it's important to still support de-serialization of legacy. I think we should default to the new though as that will be the default moving forward - so swap `cls._signed_transaction_serializer` as the first attempt. Thoughts?

eth_account/typed_transactions/blob_transactions/blob_transaction.py
```diff
@@ -250,17 +280,25 @@ def from_bytes(cls, encoded_transaction: HexBytes) -> "BlobTransaction":
...
+                dictionary = cls._legacy_signed_pooled_transaction_serializer.from_bytes(  # type: ignore  # noqa: E501
+                    transaction_payload
+                ).as_dict()
+                # Legacy format has 'proofs' instead of 'cell_proofs', but we don't
+                # need to convert since cell_proofs will be recomputed from blobs
+            except rlp.exceptions.ObjectDeserializationError:
+                # Fall back to transaction without blob data
```
fselmo, line 292: I'm not sure what this comment means. I would prefer:


```suggestion
                # Fall back to EIP-4844 format (without wrapper_version, with proofs)
```

And above (where this comment would go) we can have:

```
# Try legacy EIP-7594 format (with wrapper_version and cell_proofs)

File: eth_account/typed_transactions/blob_transactions/blob_transaction.py [hunk 3]
```diff
@@ -350,8 +388,8 @@ def payload(self) -> bytes:
             rlp_serializer = self.__class__._signed_transaction_serializer
             payload = rlp.encode(rlp_serializer.from_dict(rlp_structured_dict))  # type: ignore # noqa: E501
         else:
-            # `PooledTransaction` as defined in EIP-4844
-            # rlp([tx_payload_body, blobs, commitments, proofs])
+            # `PooledTransaction` as defined in EIP-7994
```
fselmo:
```suggestion
            # `PooledTransaction` as defined in EIP-7594
```

fselmo (commented):
Just adding a few comments I noticed.

I'm not sure about the go code approach here. We already have a tested implementation as an import here with `ckzg`. I think we need to just make sure we are implementing the transaction part correctly so I think it would be sufficient to point to test cases somewhere where there are known resulting values and make sure we write pytest tests that come up with the same results, as we did before.

Thoughts on this?

eth_account/typed_transactions/blob_transactions/blob_transaction.py
fselmo: Ah I was reading this wrong, sorry. Please ignore this comment :)

eth_account/typed_transactions/blob_transactions/blob_transaction.py
fselmo: ignore :)

tests/core/test_blob_transactions.py
```diff
@@ -90,67 +104,38 @@ def test_blobs_commitments_proofs_and_hashes_from_blobs():
     assert tx.blob_data.proofs[0].as_hexstr() == ZERO_BLOB_COMMITMENT_AND_PROOF_HASH
 
 
-def test_sign_blob_transaction_with_zero_blob_and_compare_with_tx_from_bytes():
-    # test case taken from web3j library:
-    # https://github.com/web3j/web3j/blob/c0679c08dfc1e1b196d1086f47d2eedc72f2a8f7/crypto/src/test/resources/blob_data/blob_tx_signed.txt  # noqa: E501
-    # https://github.com/web3j/web3j/blob/c0679c08dfc1e1b196d1086f47d2eedc72f2a8f7/crypto/src/test/java/org/web3j/crypto/TransactionEncoderTest.java#L123-L128  # noqa: E501
-    with open(SIGNED_TX_PATH) as signed_tx_file:
-        signed_tx_from_file = to_bytes(hexstr=signed_tx_file.read().strip("\n"))
+def test_sign_blob_transaction_matches_expected_bytes():
```
fselmo, line 107: we should keep `zero_blob` in this test name, no?

tests/core/test_blob_transactions.py
```diff
@@ -163,3 +148,160 @@ def test_high_and_low_blob_count_limit_validation():
...
+        == ZERO_BLOB_COMMITMENT_AND_PROOF_HASH
+    )
+
+    for _i, (original, deserialized) in enumerate(
+        zip(tx.blob_data.cell_proofs, tx_from_bytes.blob_data.cell_proofs)
+    ):
+        assert original.as_hexstr() == deserialized.as_hexstr()
+
+
+def _get_go_eth_kzg_test_cases():
```
fselmo, line 277: We can use a docstring here for a reader. This isn't immediately obvious what it's doing. IIUC, we are reading some pre-computed go-eth-kzg results for the same test cases as the consensus-spec-test cases and then building a more readily parseable set of test cases to compare our implementation against go-eth-kzg using the test cases from consensus specs?

We could also maybe use an `__init__.py` or a `README.md` inside of `_test_data/` that better describes what was used to generate the `go_eth_zkg_computed`. I'm certain the same exact tx parameters had to be used, etc. It might just be good to jot this down somewhere. Maybe this is in the docstring of the actual test itself and not at the module level but I think it would be good to have this information.

Even if you removed the script that produced this, it might be useful to link to the commit that was used to generate it too in case someone wants to look at it and we should rebase merge this to keep that history preserved for debugging - just make sure to manually squash things into cleaner commits.

Thoughts on the above? This is really great though, really nice testing 👌🏼

fselmo (commented):
> I'm also curious whether or not you think we should return 4844 values if they get passed in, or if we should accept both and always return the 7594 format?

I'm not sure what you mean by return here. If we have a signed 4844 transaction that we de-serialize as a `BlobTransaction`, then we can calculate any of the crypto fields as `@computed_field` so you can get any of that information from the blob transaction.

I'm not sure why we don't have a `as_signed_transaction()` method in the case that we have all the fields tbh... in which case it would determine which serializer we use. I supposer the `payload()` method uses cell proofs in this sense (7594) and I think this is appropriate. Are you wondering if we should have a way to sign it as either / or?

tests/core/test_blob_transactions.py
kclowes: Sounds good. I will add some documentation. I'll also put up the script again, and you can decide if you want to keep it in the codebase or not. It was mostly claude-generated so I think it's fairly sloppy but feel free to do whatever makes more sense!

kclowes:
Okay, I think this is ready. Feel free to update and merge as you wish! I did not get a chance to test out this API via web3 yet, but I think as long as the blob txs were working, this should too and now we're in beta, so you should be fine.

fselmo:
> I did not get a chance to test out this API via web3 yet

I also checked this locally and it looks good 👍🏼

PR #328, opened 2025-12-12
Template upgrade

kclowes:
### What was wrong?
Template updates! Mainly drops Python 3.8 and 3.9 support and adds support for Python 3.14.

### How was it fixed?
Pulled in the template, ran the linter. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->:max_bytes(150000):strip_icc()/__opt__aboutcom__coeus__resources__content_migration__mnn__images__2015__09__sea-otter-fur-dense-b4eecf1a0729431c97609845d93ff002.jpg)


eth_account/types.py
```diff
@@ -23,8 +23,14 @@
...
-AccessList = Sequence[Dict[str, Union[HexStr, Sequence[HexStr]]]]
-RLPStructuredAccessList = Sequence[Sequence[Union[HexStr, Sequence[HexStr]]]]
+
+class AccessListEntry(TypedDict):
+    address: HexStr
+    storageKeys: Sequence[HexStr]
+
+
+AccessList = Sequence[AccessListEntry]
+RLPStructuredAccessList = Sequence[tuple[HexStr, Sequence[HexStr]]]
```
kclowes, line 33: The latest mypy was choking on the old typing because of the `Union[HexStr, Sequence[HexStr]]`. Based on the EIP, I think the storage keys can only be a `Sequence[HexStr]`, but if that's wrong or we want to be more permissive and `cast` I can change it back. It was [this line](https://github.com/ethereum/eth-account/pull/328/files#diff-8b895c426e1ecf9ca9c22d2e8924d3b55ceda3376b2288094fb2d00bda3baa79L131) that doesn't make sense if this is allowed to be a `HexStr`, because then you're iterating over characters instead of a sequence.

fselmo (approved):

PR #333, opened 2026-02-02
Breaking/add remaining breaking changes for major

### LINKED ISSUE #296: Breaking changes for `v0.14.0`

 - The use of a `str` language arg is being deprecated in `v0.13.x`. Remove it as an option in `v0.14.0`.
 - Along with the above, the return type from `Mnemonic.detect_language` should change to a `Language` enum.

### LINKED ISSUE #305: Pin `eth-keyfile` dependency to >=0.9.0

fselmo:
### What was wrong?

We have a few more breaking changes we can make here before the next major.

- closes #296
- closes #305

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2026-02-02 at 15 04 52

### DIFF
--- a/eth_account/account.py
+++ b/eth_account/account.py
@@ -207,7 +207,7 @@ class Account(AccountLocalActions):
         # type ignored because eth_keyfile appears to be using the wrong type for
         # the password arg.
         # once fixed there, this should error and can be removed
-        return HexBytes(decode_keyfile_json(keyfile, password_bytes))  # type: ignore[arg-type]  # noqa: E501
+        return HexBytes(decode_keyfile_json(keyfile, password_bytes))  # noqa: E501
 
     @classmethod
     def encrypt(
@@ -277,7 +277,7 @@ class Account(AccountLocalActions):
         # the password arg.
         # once fixed there, this should error and can be removed
         return create_keyfile_json(
-            key_bytes, password_bytes, kdf=kdf, iterations=iterations  # type: ignore[arg-type]  # noqa: E501
+            key_bytes, password_bytes, kdf=kdf, iterations=iterations  # noqa: E501
         )
 
     @combomethod
@@ -384,7 +384,7 @@ class Account(AccountLocalActions):
         self,
         passphrase: str = "",
         num_words: int = 12,
-        language: Language | str = Language.ENGLISH,
+        language: Language = Language.ENGLISH,
         account_path: str = ETHEREUM_DEFAULT_PATH,
     ) -> tuple[LocalAccount, str]:
         r"""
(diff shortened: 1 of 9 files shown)


eth_account/account.py
```diff
@@ -207,7 +207,7 @@ def decrypt(keyfile_json: str | dict[str, Any], password: str) -> HexBytes:
         # type ignored because eth_keyfile appears to be using the wrong type for
         # the password arg.
         # once fixed there, this should error and can be removed
-        return HexBytes(decode_keyfile_json(keyfile, password_bytes))  # type: ignore[arg-type]  # noqa: E501
+        return HexBytes(decode_keyfile_json(keyfile, password_bytes))  # noqa: E501
```
kclowes: nit: format line to remove E501

eth_account/account.py
```diff
@@ -277,7 +277,7 @@ def encrypt(
         # the password arg.
         # once fixed there, this should error and can be removed
         return create_keyfile_json(
-            key_bytes, password_bytes, kdf=kdf, iterations=iterations  # type: ignore[arg-type]  # noqa: E501
+            key_bytes, password_bytes, kdf=kdf, iterations=iterations  # noqa: E501
```
kclowes: same here re. removing E501

uv.lock
kclowes, line 1: I think we can/should remove?

newsfragments/333.breaking.rst
```diff
@@ -0,0 +1 @@
+Bump ``eth-keyfile`` and ``eth-utils`` lower pins, getting ready to release next major version. Remove deprecated ``str`` support for mnemonic, opt for enum support only.
```
kclowes: I don't see the eth-utils bump, just commenting so we come back and update once eth-utils stable is out

kclowes (commented):
LGTM! Commenting rather than approving so that we remember to bump eth-utils once it's out. Let me know if you want to do that in a separate PR and I'll switch to an approval!

uv.lock
fselmo: Oops yeah, remnant of just running with `uv` everywhere nowadays. Should maybe add to `.gitignore` at some point unless unless it becomes relevant.

newsfragments/333.breaking.rst
fselmo: Right... I removed it... I think this [newsfragment] just needs an update. We already have the `CamelModel`. We decided to keep the lower pin untouched, right? 👀

fselmo: Unless we want to raise it at least here and in `web3.py` so we can use new features within a release cycle (I'd prefer this). If we plan this, I think we should lower pin here to the beta before releasing so we can sanity check there are no issues? 👀

fselmo: ^^ as discussed... will bump when stable is out in separate PR

kclowes (approved):
🚀 !

